PR #16, opened 2023-07-13
Add hydra service

nsklikas:
IAM-363: hydra oauth2 endpoints


cmd/main.go
```diff
@@ -32,7 +33,9 @@ func main() {
 	monitor := prometheus.NewMonitor("identity-admin-ui", logger)
 	tracer := tracing.NewTracer(tracing.NewConfig(specs.TracingEnabled, specs.JaegerEndpoint, logger))
 
-	router := web.NewRouter(tracer, monitor, logger)
+	hClient := ih.NewClient(specs.HydraAdminURL)
```
shipperizer, line 36: small tiny bit here (which u haven't been exposed before)

we gotta do a check on this env config var, if empty all falls apart, where we do it can differ:

* use annotations (should be doable) and make it not nullable
* simply `if specs.HydraAdminURL ==  "" {panic("empty url")}`
* panic inside the `ih.NewClient`

basically panic and fail hard and fast

internal/hydra/client.go
```diff
@@ -0,0 +1,30 @@
...
+func (c *Client) OAuth2Api() client.OAuth2Api {
+	return c.c.OAuth2Api
+}
+
+func NewClient(url string) *Client {
+	c := new(Client)
+
+	configuration := client.NewConfiguration()
+
+	configuration.Debug = true
```
shipperizer, line 20:
```suggestion
        // TODO @nsklikas pass this in as arg
	configuration.Debug = true
```

internal/responses/responses.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func NewResponse(data interface{}, msg string, status int, meta interface{}) *Response {
+	r := new(Response)
+	r.Data = data
+	r.Message = msg
+	r.Status = status
+	r.Meta = meta
+	return r
+}
```
shipperizer, line 27: we don't really add anything to the table

```
r := new(response.Response)

r.Data = data
r.Message = msg
r.Status = status
r.Meta = meta

rrr, err := json.Marshal(r)
if err != nil {
               w.WriteHeader(http.ServerInternalError)
	      return 
}

w.Write(rr)
w.WriteHeader(http.StatusOK)

```

logic above is simple enough to be in the handler imho, hiding it behind the PrepareResponse simply adds more `ifs` and hids it , arguable point i know, i would stick with explicit for now

pkg/clients/service.go
```diff
@@ -0,0 +1,72 @@
...
+
+type Service struct {
+	hydra HydraClientInterface
+
+	tracer  trace.Tracer
+	monitor monitoring.MonitorInterface
+	logger  logging.LoggerInterface
+}
+
+func (s *Service) parseServiceError(r *http.Response, e error) *hClient.ErrorOAuth2 {
```
shipperizer, line 23: took me a while to understand that the hydra client doesn't return error objects, mind adding a small description or changing the name of the method?

pkg/clients/service.go
```diff
@@ -0,0 +1,72 @@
...
+	json_data, err := io.ReadAll(data)
+	if err != nil {
+		return err
+	}
+	err = json.Unmarshal(json_data, v)
+	if err != nil {
+		return err
+	}
+	return nil
+}
```
shipperizer, line 72: i'd stay away from the `any` generic, better to put `interface{}` or probably do this inside the function inline so that it doesn't feel like magic

pkg/clients/handlers.go
```diff
@@ -0,0 +1,58 @@
...
+	w.Header().Set("Content-Type", "application/json")
+	rr, err := r.PrepareResponse()
+	if err != nil {
+		a.logger.Errorf("Unexpected error: %s", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.Write(rr)
+	w.WriteHeader(r.Status)
+}
```
shipperizer, line 31: i have some reserves in this, but i'd say crack on with the next 3 HTTP methods and see where u get with this and if it stays the same

pkg/clients/handlers.go
```diff
@@ -0,0 +1,166 @@
...
+
+	rr, err := json.Marshal(r)
+	if err != nil {
+		a.logger.Errorf("Unexpected error: %s", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.Write(rr)
+	w.WriteHeader(r.Status)
+}
```
shipperizer, line 46: still not convinced about this, but right now cannot find a better - alternative without poking around further

pkg/clients/handlers.go
```diff
@@ -0,0 +1,166 @@
...
+	c, errResp, e := a.service.CreateClient(r.Context(), c)
+	if errResp != nil {
+		a.WriteJSONResponse(w, errResp, "", int(*errResp.StatusCode), nil, nil)
+		return
+	}
+	if e != nil {
+		a.logger.Errorf("Unexpected error: %s", e)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
```
shipperizer, line 89: as discussed, there might be a way to offload all of this to the service layer, simply passing the context and the unmarshalled/deserialezed body 

see if there is a nicer way, if not this does the job

pkg/clients/service.go
```diff
@@ -0,0 +1,196 @@
...
+		Execute()
+
+	if err != nil {
+		se, err := s.parseServiceError(resp, err)
+		return nil, se, err
+	}
+	return c, nil, nil
+}
+
+func (s *Service) ListClients(ctx context.Context, cName string, owner string, ps int64, pt string, u *url.URL) ([]hClient.OAuth2Client, *PaginationMetadata, *hClient.ErrorOAuth2, error) {
```
shipperizer, line 114: I think a pagination struct composed with a specific service pagination struct would work better here instead of all the arts (see golang composition)


```suggestion
func (s *Service) ListClients(ctx context.Context, cName, owner, pt string, ps int64, u *url.URL) ([]hClient.OAuth2Client, *PaginationMetadata, *hClient.ErrorOAuth2, error) {
```

pkg/clients/service.go
```diff
@@ -0,0 +1,196 @@
...
+	}
+	if owner != "" {
+		r = r.Owner(owner)
+	}
+	if ps != 0 {
+		r = r.PageSize(ps)
+	}
+	if pt != "" {
+		r = r.PageToken(pt)
+	}
```
shipperizer, line 130: if u add the struct u can spare yourself from these checks in here

pkg/clients/service.go
```diff
@@ -0,0 +1,196 @@
...
+	return p
+}
+
+func (s *Service) parseLink(l, t string, u *url.URL) string {
+	// We extract the link for inputs of the format:
+	// '<{link}>; rel="first",<{link}>; rel="prev",<{link}>; rel="next",<{link}>; rel="last"'
+	// Where t is one of ["first", "prev", "next", "last"] (not all of them need to exist)
+	// TODO @nsklikas: evaluate performance cost? Is there better way to do this?
+	re := regexp.MustCompile("<[^>]*>; rel=\"" + t + "\"")
+	ree := regexp.MustCompile("<[^>]*>")
```
shipperizer, line 166: chuck them in the NewService func, constructor material

pkg/clients/handlers.go
```diff
@@ -0,0 +1,166 @@
...
+	service ServiceInterface
+
+	logger logging.LoggerInterface
+}
+
+func (a *API) RegisterEndpoints(mux *chi.Mux) {
+	mux.Get("/clients", a.ListClients)
+	mux.Post("/clients", a.CreateClient)
+	mux.Get("/clients/{client_id}", a.GetClient)
+	mux.Put("/clients/{client_id}", a.UpdateClient)
```
shipperizer, line 25: prefix everything with `/api/v0/`

pkg/oauth_clients/handlers.go
```diff
@@ -0,0 +1,62 @@
...
+)
+
+type API struct {
+	service ServiceInterface
+
+	logger logging.LoggerInterface
+}
+
+func (a *API) RegisterEndpoints(mux *chi.Mux) {
+	mux.Get("/clients/{client_id}", a.ConstructHandler(a.GetClient))
```
shipperizer, line 17: uri needs `/api/v0` prefix on top of needing a different uri

pkg/oauth_clients/interfaces.go
```diff
@@ -0,0 +1,27 @@
+package oauth_clients
```
shipperizer, line 1: unless it's a double copy of the previous package or a wip, not sure this should be here?

nsklikas: deleted

pkg/clients/service.go
```diff
@@ -0,0 +1,280 @@
...
+			return nil, err
+		}
+		ret.ServiceError = se
+	}
+	ret.Resp = c
+	return ret, nil
+}
+
+func (s *Service) DeleteClient(ctx context.Context, clientID string) (*ServiceResponse, error) {
+	_, span := s.tracer.Start(ctx, "hydra.OAuth2Api.DeleteOAuth2Client")
```
shipperizer, line 81: sorry i ve been missing a trick here, basically the context needs to be overridden (enhanced) and passed onwards so traces are propagated

```suggestion
	ctx, span := s.tracer.Start(ctx, "hydra.OAuth2Api.DeleteOAuth2Client")
```

this should be applied in every `tracer.Start`

pkg/clients/handlers.go
```diff
@@ -0,0 +1,224 @@
...
+	r.Meta = meta
+
+	rr, err := json.Marshal(r)
+	if err != nil {
+		a.logger.Errorf("Unexpected error: %s", err)
+		w.WriteHeader(http.StatusInternalServerError)
+		return
+	}
+	w.WriteHeader(status)
+	w.Write(rr)
```
shipperizer, line 53: u can do this or u can try 

```
	json.NewEncoder(w).Encode(
		<response struct with json annotations>,
	)
	w.WriteHeader(http.StatusOK)
```

pkg/clients/handlers.go
```diff
@@ -0,0 +1,224 @@
...
+	Prev  string `json:"prev,omitempty"`
+	Next  string `json:"next,omitempty"`
+}
+
+func (a *API) RegisterEndpoints(mux *chi.Mux) {
+	mux.Get("/api/v0/clients", a.ListClients)
+	mux.Post("/api/v0/clients", a.CreateClient)
+	mux.Get("/api/v0/clients/{client_id}", a.GetClient)
+	mux.Put("/api/v0/clients/{client_id}", a.UpdateClient)
+	mux.Delete("/api/v0/clients/{client_id}", a.DeleteClient)
```
shipperizer, line 33: minor thing, but here i'd stick with `{id}` reason is it is simple to do regexp patterns in the middleware when the `id` name is standardized

shipperizer (approved):
:shipit

as discussed let's review the common response approach once IAM-364 is done

PR #20, opened 2023-08-09
IAM 365

shipperizer:
IAM-365: idp APIs

- feat: add idp service
- feat: add idp handlers
- ci: local deployment setup for k8s via skaffold
- feat: wire up main and router with new dependencies
- fix: use new instead of & syntax
- fix: add otel tracing to hydra client
- fix: create k8s coreV1 package
- fix: use io pkg instead of ioutil
- fix: drop unused const


pkg/idp/service.go
```diff
@@ -0,0 +1,327 @@
...
+	}
+
+	var idp *Configuration
+	// TODO @shipperizer find a better way to index the idps
+	for _, i := range idps {
+		if i.ID == providerID {
+			i = s.mergeConfiguration(i, data)
+			idp = i
+		}
+	}
```
nsklikas, line 101: You could use the provider ID as the key in the configMap, you can switch from:
```yaml
idps.yaml:
  - provider_id: provider_1
    ...
  - provider_id: provider_2
    ...
  - provider_id: provider_3
    ...
```
To:
```yaml
provider_1:
  provider_id: provider_1
  ...
provider_2:
  provider_id: provider_2
  ...
provider_3:
  provider_id: provider_3
  ...
```
This way you should be able to fetch the provider simply by indexing the map. Is there a problem with this?

nsklikas (approved):

PR #63, opened 2023-08-25
fix: add default schema changes

shipperizer:


nsklikas:
This design looks fine for me. 

One thought:
We could refactor the cm to have the following structure:
```
default_schema_id: <the default schema ID>
schemas: 
    <schema_id>: <schema>
```
This would mean to "push" the current mapping of `schema_id: schema` from the top level to under the the `schemas` key. This would separate the `default_schema_id` from the schema mapping, making it easier to parse it and it would simplify edge cases with managing the cm (e.g. what if schema_id == "default_schema_id"?)

PR #84, opened 2023-09-06
Fix renovate workflow

nsklikas:
IAM-414


.github/.jira_sync_config.yaml
```diff
@@ -0,0 +1,37 @@
...
+
+  # Dictionary mapping GitHub issue status to Jira issue status
+  status_mapping:
+    opened: Untriaged
+    closed: done
+
+  # (Optional) Jira project components that should be attached to the created issue
+  # Component names are case-sensitive
+  components:
+    - Kratos
```
natalian98, line 13:
```suggestion
    - Login UI
```

.github/.jira_sync_config.yaml
```diff
@@ -0,0 +1,37 @@
...
+  add_gh_comment: false
+
+  # (Optional) (Default: true) Synchronize issue description from GitHub to Jira
+  sync_description: true
+
+  # (Optional) (Default: true) Synchronize comments from GitHub to Jira
+  sync_comments: true
+
+  # (Optional) (Default: None) Parent Epic key to link the issue to
+  epic_key: "IAM-19"
```
natalian98, line 31: Maybe the issues should be linked to Login UI epic instead?

.github/.jira_sync_config.yaml
nsklikas: That's from another branch, it was just showing because I force pushed main and didnt rebase this branch. It's gone now.

.github/.jira_sync_config.yaml
nsklikas: See https://github.com/canonical/identity-platform-admin-ui/pull/93

.github/workflows/auto-approver.yaml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - uses: actions/checkout@f43a0e5ff2bd294095638e18286ca9a3d1956744 # v3
+      - name: Approve PR
+        run: |
+          gh pr review --approve || true
+        env:
+          GITHUB_TOKEN: ${{ secrets.PAT_TOKEN }}
+      - name: Enable automerge if required
+        run: |
+          if [[ $(gh pr view --json body | grep "\*\*Automerge\*\*: Enabled") ]]; then
```
shipperizer, line 21: https://docs.renovatebot.com/configuration-options/#additionalbranchprefix

what about using that in the config, have a different prefix per type of dependency stuff and then use the `if branch name == x` in the github action yaml?

basically parsing the body seems a bit risky

nsklikas: Updated, should be better now

shipperizer (approved):

PR #133, opened 2023-10-09
IAM 282

shipperizer:
- ci: IAM-282 introduce structural tests
- fix: use version from release-please worflow
- fix: introduce version flag to facilitate charm code
- fix: use version in /api/v0/version endpoint


internal/version/const.go
```diff
@@ -0,0 +1,3 @@
+package version
+
+const Version = "v1.2.0" // x-release-please-version
```
nsklikas, line 3:
```suggestion
const Version = "1.2.0" // x-release-please-version
```

nsklikas (approved):

PR #157, opened 2023-10-20
IAM-337-Implement Oathkeeper integration

bencekov:
This PR includes the implementation of the Admin UI Rest interface to access and manipulate Oathkeeper access rules.

Testing Guide:

The following endpoints have been implemented for Admin UI:

Get /api/v0/rules
Get /api/v0/rules/{id}
Post /api/v0/rules
Put /api/v0/rules
Delete /api/v0/rules/{id}

{id} is a URL parameter that accepts the Oathkeeper rule IDs specified in a Kubernetes configmap.

The make command make dev sets up a development environment.

### DIFF
--- a/cmd/main.go
+++ b/cmd/main.go
@@ -11,16 +11,18 @@ import (
 	"time"
 
 	ih "github.com/canonical/identity-platform-admin-ui/internal/hydra"
-	"github.com/canonical/identity-platform-admin-ui/internal/k8s"
+	k8s "github.com/canonical/identity-platform-admin-ui/internal/k8s"
 	"github.com/kelseyhightower/envconfig"
 
 	"github.com/canonical/identity-platform-admin-ui/internal/config"
 	ik "github.com/canonical/identity-platform-admin-ui/internal/kratos"
 	"github.com/canonical/identity-platform-admin-ui/internal/logging"
 	"github.com/canonical/identity-platform-admin-ui/internal/monitoring/prometheus"
+	io "github.com/canonical/identity-platform-admin-ui/internal/oathkeeper"
 	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
 	"github.com/canonical/identity-platform-admin-ui/internal/version"
 	"github.com/canonical/identity-platform-admin-ui/pkg/idp"
+	"github.com/canonical/identity-platform-admin-ui/pkg/rules"
 	"github.com/canonical/identity-platform-admin-ui/pkg/schemas"
 	"github.com/canonical/identity-platform-admin-ui/pkg/web"
 )
@@ -51,6 +53,7 @@ func main() {
 	hAdminClient := ih.NewClient(specs.HydraAdminURL, specs.Debug)
 	kAdminClient := ik.NewClient(specs.KratosAdminURL, specs.Debug)
 	kPublicClient := ik.NewClient(specs.KratosPublicURL, specs.Debug)
+	oPublicClient := io.NewClient(specs.OathkeeperPublicURL, specs.Debug)
 
 	k8sCoreV1, err := k8s.NewCoreV1Client()
 
@@ -70,7 +73,10 @@ func main() {
 		Name:      specs.SchemasConfigMapName,
 		Namespace: specs.SchemasConfigMapNamespace,
 	}
-	router := web.NewRouter(idpConfig, schemasConfig, hAdminClient, kAdminClient, tracer, monitor, logger)
+
+	rulesConfig := rules.NewConfig(specs.RulesConfigMapName, specs.RulesConfigMapNamespace, k8sCoreV1, oPublicClient.ApiApi())
+
+	router := web.NewRouter(idpConfig, schemasConfig, rulesConfig, hAdminClient, kAdminClient, tracer, monitor, logger)
 
 	logger.Infof("Starting server on port %v", specs.Port)
 
--- a/internal/config/specs.go
+++ b/internal/config/specs.go
@@ -13,13 +13,17 @@ type EnvSpec struct {
 
 	Debug bool `envconfig:"debug" default:"false"`
 
-	KratosPublicURL string `envconfig:"kratos_public_url" required:"true"`
-	KratosAdminURL  string `envconfig:"kratos_admin_url" required:"true"`
-	HydraAdminURL   string `envconfig:"hydra_admin_url" required:"true"`
+	KratosPublicURL     string `envconfig:"kratos_public_url" required:"true"`
+	KratosAdminURL      string `envconfig:"kratos_admin_url" required:"true"`
+	HydraAdminURL       string `envconfig:"hydra_admin_url" required:"true"`
+	OathkeeperPublicURL string `envconfig:"oathkeeper_public_url" required:"true"`
 
 	IDPConfigMapName      string `envconfig:"idp_configmap_name" required:"true"`
 	IDPConfigMapNamespace string `envconfig:"idp_configmap_namespace" required:"true"`
 
 	SchemasConfigMapName      string `envconfig:"schemas_configmap_name" required:"true"`
 	SchemasConfigMapNamespace string `envconfig:"schemas_configmap_namespace" required:"true"`
+
+	RulesConfigMapName      string `envconfig:"rules_configmap_name" required:"true"`
+	RulesConfigMapNamespace string `envconfig:"rules_configmap_namespace" required:"true"`
 }
--- a/pkg/rules/handlers.go
+++ b/pkg/rules/handlers.go
@@ -1,0 +1,241 @@
+package rules
+
+import (
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+
+	"github.com/canonical/identity-platform-admin-ui/internal/http/types"
+	"github.com/canonical/identity-platform-admin-ui/internal/logging"
+
+	"github.com/go-chi/chi/v5"
+	oathkeeper "github.com/ory/oathkeeper-client-go"
+)
+
+type API struct {
+	service ServiceInterface
+
+	logger logging.LoggerInterface
+}
+
+func (a *API) RegisterEndpoints(mux *chi.Mux) {
+	mux.Get("/api/v0/rules", a.handleList)
+	mux.Get("/api/v0/rules/{id}", a.handleDetail)
+	mux.Post("/api/v0/rules", a.handleCreate)
+	mux.Put("/api/v0/rules/{id}", a.handleUpdate)
+	mux.Delete("/api/v0/rules/{id}", a.handleRemove)
+}
+
+func (a *API) handleList(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	pagination := types.ParsePagination(r.URL.Query())
+
+	rules, err := a.service.ListRules(r.Context(), pagination.Page, pagination.Size)
+
+	if err != nil {
+
+		rr := types.Response{
+			Status:  http.StatusInternalServerError,
+			Message: err.Error(),
+		}
+
+		w.WriteHeader(http.StatusInternalServerError)
+		json.NewEncoder(w).Encode(rr)
+
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	json.NewEncoder(w).Encode(
+		types.Response{
+			Data:    rules,
+			Message: "List of rules",
+			Status:  http.StatusOK,
+		},
+	)
+}
+
+func (a *API) handleDetail(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	ruleID := chi.URLParam(r, "id")
+
+	rule, err := a.service.GetRule(r.Context(), ruleID)
+
+	if err != nil {
+
+		rr := types.Response{
+			Status:  http.StatusInternalServerError,
+			Message: err.Error(),
+		}
+
+		w.WriteHeader(http.StatusInternalServerError)
+		json.NewEncoder(w).Encode(rr)
+
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	json.NewEncoder(w).Encode(
+		types.Response{
+			Data:    rule,
+			Message: "Rule detail",
+			Status:  http.StatusOK,
+		},
+	)
+}
+
+func (a *API) handleCreate(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	defer r.Body.Close()
+	body, err := io.ReadAll(r.Body)
+
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		json.NewEncoder(w).Encode(
+			types.Response{
+				Message: "Error parsing request payload",
+				Status:  http.StatusBadRequest,
+			},
+		)
+
+		return
+	}
+
+	rule := new(oathkeeper.Rule)
+	if err := json.Unmarshal(body, rule); err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		json.NewEncoder(w).Encode(
+			types.Response{
+				Message: "Error parsing JSON payload",
+				Status:  http.StatusBadRequest,
+			},
+		)
+
+		return
+
+	}
+
+	err = a.service.CreateRule(r.Context(), *rule)
+
+	if err != nil {
+
+		rr := types.Response{
+			Status:  http.StatusInternalServerError,
+			Message: err.Error(),
+		}
+
+		w.WriteHeader(http.StatusInternalServerError)
+		json.NewEncoder(w).Encode(rr)
+
+		return
+	}
+
+	w.WriteHeader(http.StatusCreated)
+	json.NewEncoder(w).Encode(
+		types.Response{
+			Message: fmt.Sprintf("Created rule %s", *rule.Id),
+			Status:  http.StatusCreated,
+		},
+	)
+}
+
+func (a *API) handleUpdate(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	ruleID := chi.URLParam(r, "id")
+
+	defer r.Body.Close()
+	body, err := io.ReadAll(r.Body)
+
+	if err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		json.NewEncoder(w).Encode(
+			types.Response{
+				Message: "Error parsing request payload",
+				Status:  http.StatusBadRequest,
+			},
+		)
+
+		return
+	}
+
+	rule := new(oathkeeper.Rule)
+	if err := json.Unmarshal(body, rule); err != nil {
+		w.WriteHeader(http.StatusBadRequest)
+		json.NewEncoder(w).Encode(
+			types.Response{
+				Message: "Error parsing JSON payload",
+				Status:  http.StatusBadRequest,
+			},
+		)
+
+		return
+
+	}
+
+	err = a.service.UpdateRule(r.Context(), ruleID, *rule)
+
+	if err != nil {
+
+		rr := types.Response{
+			Status:  http.StatusInternalServerError,
+			Message: err.Error(),
+		}
+
+		w.WriteHeader(http.StatusInternalServerError)
+		json.NewEncoder(w).Encode(rr)
+
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	json.NewEncoder(w).Encode(
+		types.Response{
+			Message: fmt.Sprintf("Updated rule %s", *rule.Id),
+			Status:  http.StatusOK,
+		},
+	)
+}
+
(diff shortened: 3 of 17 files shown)


internal/config/specs.go
```diff
@@ -13,13 +13,18 @@ type EnvSpec struct {
...
 
 	IDPConfigMapName      string `envconfig:"idp_configmap_name" required:"true"`
 	IDPConfigMapNamespace string `envconfig:"idp_configmap_namespace" required:"true"`
 
 	SchemasConfigMapName      string `envconfig:"schemas_configmap_name" required:"true"`
 	SchemasConfigMapNamespace string `envconfig:"schemas_configmap_namespace" required:"true"`
+
+	RulesConfigMapName      string `envconfig:"rules_configmap_name" required:"true"`
+	RulesConfigMapNamespace string `envconfig:"rules_configmap_namespace" required:"true"`
+	RulesFilesLocations     string `envconfig:"rules_file_locations" required:"true"`
```
shipperizer, line 29:
```suggestion
	RulesFilesLocations     []string `envconfig:"rules_file_locations" required:"true"`
```

pkg/rules/handlers.go
```diff
@@ -0,0 +1,267 @@
...
+	mux.Get("/api/v0/rules/{id}", a.handleDetail)
+	mux.Post("/api/v0/rules/{id}", a.handleCreate)
+	mux.Put("/api/v0/rules/{id}", a.handleUpdate)
+	mux.Delete("/api/v0/rules/{id}", a.handleRemove)
+}
+
+func (a *API) handleList(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	limit, err := strconv.ParseInt(r.URL.Query().Get("limit"), 10, 64)
```
shipperizer, line 33: look for the pagination in here https://github.com/canonical/identity-platform-admin-ui/blob/main/internal/http/types/generic.go

pkg/rules/handlers.go
```diff
@@ -0,0 +1,267 @@
...
+	return strings.Map(
+		func(r rune) rune {
+			if r == '\\' {
+				return -1
+			}
+			return r
+		},
+		s,
+	)
+}
```
shipperizer, line 267: not sure I follow this?

pkg/rules/handlers.go
```diff
@@ -0,0 +1,267 @@
...
+type API struct {
+	service ServiceInterface
+
+	logger logging.LoggerInterface
+}
+
+func (a *API) RegisterEndpoints(mux *chi.Mux) {
+	mux.Get("/api/v0/rules", a.handleList)
+	mux.Get("/api/v0/rules/{id}", a.handleDetail)
+	mux.Post("/api/v0/rules/{id}", a.handleCreate)
```
shipperizer, line 25: ID shouldn't be picked by the URL, it comes either from the payload (not great) or we create a random one from the data we ve been given, like a uuidv5

pkg/rules/handlers.go
```diff
@@ -0,0 +1,267 @@
...
+		},
+	)
+}
+
+func (a *API) handleCreate(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	ruleID := chi.URLParam(r, "id")
+	ruleID = cleanStr(ruleID)
+	fileName := r.URL.Query().Get("filename")
```
shipperizer, line 105: this needs to come from the payload somehow

pkg/rules/handlers.go
```diff
@@ -0,0 +1,267 @@
...
+		w.WriteHeader(http.StatusInternalServerError)
+		json.NewEncoder(w).Encode(rr)
+
+		return
+	}
+
+	w.WriteHeader(http.StatusOK)
+	json.NewEncoder(w).Encode(
+		types.Response{
+			Data:    rule,
```
shipperizer, line 93: return a list of 1 rule

pkg/rules/interfaces.go
```diff
@@ -0,0 +1,13 @@
...
+
+import (
+	"context"
+)
+
+type ServiceInterface interface {
+	ListRules(context.Context, int64, int64) ([]*RuleData, error)
+	GetRule(context.Context, string) (*RuleData, error)
+	UpdateRule(context.Context, string, RuleData) error
+	CreateRule(context.Context, string, RuleData, string) error
```
shipperizer, line 11: apart from reordering the args we need to drop the filename location

pkg/rules/service.go
```diff
@@ -0,0 +1,347 @@
...
+
+	var rules []*RuleData
+
+	_, ok := cm.Data[userFile]
+	if ok {
+		rules = s.extractRules(cm.Data, userFile)
+		rules = append(rules, &newRule)
+	} else {
+		rules = []*RuleData{&newRule}
+	}
```
shipperizer, line 209:
```suggestion
	if _, ok := cm.Data[userFile]; ok {
		rules = s.extractRules(cm.Data, userFile)
		rules = append(rules, &newRule)
	} else {
		rules = []*RuleData{&newRule}
	}
```

pkg/rules/service.go
```diff
@@ -0,0 +1,347 @@
...
+		newRules := make([]*RuleData, 0)
+
+		for _, r := range rules {
+
+			if r.Id == id {
+				found = true
+			} else {
+				newRules = append(newRules, r)
+			}
+		}
```
shipperizer, line 269: probably easier to pop the element than re adding all of the ones needed

pkg/rules/service.go
```diff
@@ -0,0 +1,347 @@
...
+			return err
+		}
+
+		return nil
+	}
+
+	return fmt.Errorf("rule with ID %s not found", id)
+}
+
+func (s *Service) extractRules(data map[string]string, fileName string) []*RuleData {
```
shipperizer, line 294: you might need to change data structure to perform operations in a more efficient way, like using a map of ruledata

pkg/rules/service.go
```diff
@@ -0,0 +1,347 @@
...
+
+	s.oathkeeper = oClient
+	s.k8s = config.K8s
+	s.cmName = config.Name
+	s.cmNamespace = config.Namespace
+
+	if config.RuleFiles == "" {
+		s.ruleFiles = nil
+	} else {
+		s.ruleFiles = strings.Split(config.RuleFiles, ";")
```
shipperizer, line 330: use envconfig []string if possible

internal/config/specs.go
bencekov: how do you pass a list as an environment variable?

pkg/rules/handlers.go
bencekov: The examples I've seen for oathkeeper rules had IDs that included a :. Because of this I thought that some IDs that were parameters of the URL would include characters that might be escaped. I wrote this function to check for escaped characters in the Id parameters. I agree with your other point that having the ID as a url parameter is misguided, so I'll take this part out.

pkg/rules/handlers.go
bencekov: sure

pkg/rules/handlers.go
bencekov: I've seen only a couple of examples for oathkeeper access rules. The convention for the ID (for example: zinc-k8s:deny) seems to be closely tied to the app and the function of the rule. I'll check with Natalia. If it's unconventional to have random ids than I would rather not implement it, but I'll definitely take the ID url parameter out.

internal/config/specs.go
shipperizer: same way as you are doing manually now, splitting by commas https://github.com/kelseyhightower/envconfig/blob/v1.4.0/envconfig.go

check the docs for better examples

pkg/rules/service.go
```diff
@@ -0,0 +1,347 @@
...
+			return err
+		}
+
+		return nil
+	}
+
+	return fmt.Errorf("rule with ID %s not found", id)
+}
+
+func (s *Service) CreateRule(ctx context.Context, id string, newRule RuleData, userFile string) error {
```
nsklikas, line 160: I think that it would make our life easier if we created a different file for each user created rule. That way we wouldn't have to iterate over all the files every time. 

We can discuss it after standup

pkg/rules/service.go
```diff
@@ -0,0 +1,347 @@
...
+
+	if _, err = s.k8s.ConfigMaps(s.cmNamespace).Update(ctx, cm, metaV1.UpdateOptions{}); err != nil {
+
+		return err
+	}
+
+	return nil
+}
+
+func (s *Service) DeleteRule(ctx context.Context, id string) error {
```
nsklikas, line 227: What happens if a user tries to delete a rule that was created from an integration?? If I understand correctly the rule will be deleted, but it will later be re-created. 

We probably need to return an error in that case and not delete the rule (or we need to change the oathekeeper logic).

pkg/rules/service.go
```diff
@@ -0,0 +1,347 @@
...
+		keys := make([]string, 0)
+		for k := range cm.Data {
+			keys = append(keys, k)
+		}
+		if len(keys) > 0 {
+			s.ruleFiles = keys
+		} else {
+			return fmt.Errorf("rule with ID %s not found", id)
+		}
+	}
```
nsklikas, line 250: I don't think this is needed, also AFAICT it is not correct as it assumes that only this process will be making changes to the ruleFiles.

nsklikas (changes requested):
Left some comments, we need to refactor some of the logic before we merge this. Also when taking decisions that are not obvious, try to leave a comment so that the reviewers and future devs can understand the reason behind it without having to spend too much time.

Main points:
- We need to rethink if we want to dump all user created rules to a single file
- Why do we need the conversion classes
- We need to come up with a convention for rule IDs so that we can easily deduce the file name from the rule ID.

shipperizer:
small nitpicking

there are 3 commits, first one is containing everything, second is tests and third is a "bugfix" of a thing introduced in the first commit

ideally u want to silo commits, so tests go with logic, splitting parts of the logic in concepts helps with commit siloing
if a fix is for a big introduced by the very same branch code, that is defo not a big, but just a wip

pkg/rules/service.go
bencekov: In the new code, you get an error saying that the id you gave isn't among the editable rules.

pkg/rules/handlers.go
```diff
@@ -0,0 +1,268 @@
...
+	service ServiceInterface
+
+	logger logging.LoggerInterface
+}
+
+func (a *API) RegisterEndpoints(mux *chi.Mux) {
+	mux.Get("/api/v0/rules", a.handleList)
+	mux.Get("/api/v0/rules/{id}", a.handleDetail)
+	mux.Post("/api/v0/rules", a.handleCreate)
+	mux.Put("/api/v0/rules", a.handleUpdate)
```
shipperizer, line 28: unfortunately this needs the id in the path

internal/config/specs.go
bencekov: The env var will not be used. Instead I'm using a function that is a one way conversion from an ID to a filename. All rules will have their own files. The files are prefixed by ADMIN_RULE_. Only rules originating from admin-ui can be edited.

pkg/rules/handlers.go
bencekov: This function has been taken out. Because the rule ids are also url parameters for some handlers, it's better for the ids to not have special characters.

pkg/rules/handlers.go
bencekov: No ID parameter for the post handler. The ID is taken out from the payload object.

pkg/rules/handlers.go
bencekov: The filename is now generated from the id.

pkg/rules/handlers.go
bencekov: done

pkg/rules/interfaces.go
bencekov: Interface redone. No more filename args.

pkg/rules/service.go
bencekov: The algorithm for this have been simplified. Extract rules returns a map indexed with the filename with a value of one file.

pkg/rules/service.go
bencekov: Changed return value to map[string]*oathkeeper-client.Rule.

pkg/rules/service.go
bencekov: Environment variable have been removed.

pkg/rules/handlers.go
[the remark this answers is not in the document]
bencekov: done

pkg/rules/handlers.go
```diff
@@ -0,0 +1,267 @@
...
+	a := new(API)
+
+	a.service = service
+
+	a.logger = logger
+
+	return a
+}
+
+func cleanStr(s string) string {
```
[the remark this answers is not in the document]
bencekov: removed

pkg/rules/conversion.go
```diff
@@ -0,0 +1,27 @@
+package rules
+
+type RuleData struct {
```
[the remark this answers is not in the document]
bencekov: I misinterpreted another API's code. I assumed yaml parsing was necessary. sdk struct derivatives have been removed.

pkg/rules/service.go
```diff
@@ -0,0 +1,347 @@
...
+
+	for _, fileName := range s.ruleFiles {
+		rules := s.extractRules(cm.Data, fileName)
```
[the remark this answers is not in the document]
bencekov: Done. Rules are collected in a map, and file names are created from ids.

pkg/rules/service.go
```diff
@@ -0,0 +1,347 @@
...
+		keys := make([]string, 0)
+		for k := range cm.Data {
+			keys = append(keys, k)
+		}
+		if len(keys) > 0 {
+			s.ruleFiles = keys
+		} else {
+			return fmt.Errorf("rule with ID %s not found", id)
+		}
+	}
```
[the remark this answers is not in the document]
bencekov: This section have been removed.

pkg/rules/service.go
```diff
@@ -0,0 +1,347 @@
...
+
+		if _, err = s.k8s.ConfigMaps(s.cmNamespace).Update(ctx, cm, metaV1.UpdateOptions{}); err != nil {
+
+			return err
+		}
+
+		return nil
+	}
+
+	return fmt.Errorf("rule with ID %s not found", id)
```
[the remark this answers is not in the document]
bencekov: Done.

pkg/rules/service.go
bencekov: I agree. Done.

pkg/rules/handlers.go
bencekov: Done. The query to admin-ui uses page and size parameters, but the outgoing request to oathkeeper uses limit and offset.

pkg/rules/service.go
bencekov: Done.

pkg/rules/service.go
bencekov: This code is now removed.

pkg/rules/handlers.go
bencekov: Done. In case of a mismatch between the id url parameter, and the id rule field the request returns with an error.

pkg/rules/service.go
```diff
@@ -0,0 +1,251 @@
...
+
+		}
+
+		names[*extract.Id] = true
+	}
+
+	return names
+}
+
+func NewService(oClient oathkeeper.ApiApi, config *Config, tracer trace.Tracer, monitor monitoring.MonitorInterface, logger logging.LoggerInterface) *Service {
```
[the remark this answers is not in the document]
bencekov: done

cmd/main.go
```diff
@@ -70,7 +73,14 @@ func main() {
 		Name:      specs.SchemasConfigMapName,
 		Namespace: specs.SchemasConfigMapNamespace,
 	}
-	router := web.NewRouter(idpConfig, schemasConfig, hAdminClient, kAdminClient, tracer, monitor, logger)
+
+	rulesConfig := &rules.Config{
+		K8s:       k8sCoreV1,
+		Name:      specs.RulesConfigMapName,
+		Namespace: specs.RulesConfigMapNamespace,
+	}
```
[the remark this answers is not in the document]
bencekov: done

pkg/rules/service.go
```diff
@@ -0,0 +1,251 @@
...
+
+	return file[0:len(ADMIN_UI_PREFIX)] == ADMIN_UI_PREFIX
+}
+
+func IdToFileName(id string) string {
+	filtered_name := regexp.MustCompile(`[^a-zA-Z0-9 ]+`).ReplaceAllString(id, "_")
+	return fmt.Sprintf("%s%s%s", ADMIN_UI_PREFIX, filtered_name, ".json")
+}
+
+func (s *Service) extractRules(data map[string]string) map[string]*oathkeeper.Rule {
```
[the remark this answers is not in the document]
bencekov: done

pkg/rules/service.go
```diff
@@ -0,0 +1,251 @@
...
+
+	if err != nil {
+		s.logger.Error(err.Error())
+		return err
+	}
+
+	rules := s.extractRules(cm.Data)
+	if _, ok := rules[id]; !ok {
+		return fmt.Errorf("rule with ID %s not found among editable rules", id)
+	}
```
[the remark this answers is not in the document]
bencekov: done

pkg/rules/service.go
```diff
@@ -0,0 +1,251 @@
...
```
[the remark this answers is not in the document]
bencekov: done

pkg/rules/service.go
```diff
@@ -0,0 +1,197 @@
...
+	if _, err = s.k8s.ConfigMaps(s.cmNamespace).Update(ctx, cm, metaV1.UpdateOptions{}); err != nil {
+
+		return err
+	}
+
+	return nil
+}
+
+func IdToFileName(id string) string {
+	filtered_name := regexp.MustCompile(`[^a-zA-Z0-9 ]+`).ReplaceAllString(id, "_")
```
nsklikas, line 176: Add a comment explaining what this regex is doing, also do we want to allow the space (` `) character in the rule ID and the file name?

Let's be explicit about what we allow in the rule IDs, AFAICT with this approach if someone tried to create `rule-id-1` and `rule-id_1` it would cause an error even thought they have different rule IDs

cmd/main.go
```diff
@@ -70,7 +73,15 @@ func main() {
 		Name:      specs.SchemasConfigMapName,
 		Namespace: specs.SchemasConfigMapNamespace,
 	}
-	router := web.NewRouter(idpConfig, schemasConfig, hAdminClient, kAdminClient, tracer, monitor, logger)
+
+	rulesConfig := &rules.Config{
```
shipperizer, line 77: as much as this is cumbersome, standardization and conventions are quite important

i would stick with a func called `NewConfig` even if we are not really initializing anything atm
see https://github.com/canonical/identity-platform-admin-ui/pull/157/files#diff-c444f711e9191b53952edb65bfd8c644419fc7695c62611dc0fb304b4fb197d6R51

pkg/rules/handlers.go
```diff
@@ -0,0 +1,243 @@
...
+	mux.Put("/api/v0/rules/{id}", a.handleUpdate)
+	mux.Delete("/api/v0/rules/{id}", a.handleRemove)
+}
+
+func (a *API) handleList(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	pagination := types.ParsePagination(r.URL.Query())
+	limit := pagination.Size
+	offset := (pagination.Page - 1) * pagination.Size
```
shipperizer, line 35: point of the service is to abstract and deal with parameters in its own way, so i d pass the pagination stuff as is and move this chunk in the service layer

https://github.com/canonical/identity-platform-admin-ui/pull/157/files#diff-f527fb4a79ca6df25497874fd4ddf6c1c996e72bb901367dacca55664e5ee2b7R37

bencekov: done

pkg/rules/service.go
bencekov: obsolete function

cmd/main.go
bencekov: done

nsklikas:
I think that something is broken. I run:
```
make dev
curl http://127.0.0.1:8001/api/v0/rules  -X POST -d '{"authenticators":[{"handler":"cookie_session"}],"authorizer":{"handler":"allow"},"id":"mocked_rule1","match":{"methods":["GET","POST"]},"mutators":[{"handler":"header"}],"upstream":{}}'
```

But then I don't see that rule in the configMap and no matter how many times I make the same curl, I never get an error that the rule already exists.

bencekov:
The latest commit solved the issue with make dev.

shipperizer (approved):

PR #174, opened 2023-10-26
IAM-459-Add Handlers for default schemas

bencekov:
Mostly based on PR #63 
Added unit tests for default schema functions and handlers.


pkg/schemas/service.go
```diff
@@ -223,6 +227,67 @@ func (s *Service) DeleteSchema(ctx context.Context, ID string) error {
...
+	for k := range cm.Data {
+		if k == DEFAULT_SCHEMA {
+			continue
+		}
+		schemaMap[k] = true
+	}
+	schemaMap[cm.Data[DEFAULT_SCHEMA]] = true
+	if _, ok := schemaMap[schema.ID]; !ok {
+		return nil, fmt.Errorf("schema with ID %s not available", schema.ID)
+	}
```
nsklikas, line 280: Not sure why this is needed. From what I can tell: you are copying the keys of a map to a set, then you also add the default schema ID in the set and then you check if the new default schema is in the set.

Why can't you just check if the original map contains the new default schema ID?

shipperizer:
> Mostly based on PR #63 Added unit tests for default schema functions and handlers.

if it's based on #63 u might want to rebase on top of that and get the diffs

pkg/schemas/service.go
bencekov: I can see that this is too much. The idea was to make sure that the new default schema ID is:
- either in the original map,
- or the default value of "default.schema" key in the original map for API to be idempotent
-  but the new schema ID can't be "default.schema"

I think these conditions should still be enforced.

pkg/schemas/service.go
```diff
@@ -223,6 +227,59 @@ func (s *Service) DeleteSchema(ctx context.Context, ID string) error {
...
+		return nil, fmt.Errorf("default schema %s missing", DEFAULT_SCHEMA)
+	}
+
+	defaultSchema := new(DefaultSchema)
+	defaultSchema.ID = ID
+
+	return defaultSchema, nil
+}
+
+func (s *Service) UpdateDefaultSchema(ctx context.Context, schema *DefaultSchema) (*DefaultSchema, error) {
```
nsklikas, line 253: I would say to rename this var, so that it is clear what it is.
```suggestion
func (s *Service) UpdateDefaultSchema(ctx context.Context, schemaID *DefaultSchema) (*DefaultSchema, error) {
```

pkg/schemas/service.go
```diff
@@ -223,6 +227,59 @@ func (s *Service) DeleteSchema(ctx context.Context, ID string) error {
...
+	cm, err := s.k8s.ConfigMaps(s.cmNamespace).Get(ctx, s.cmName, metaV1.GetOptions{})
+
+	if err != nil {
+		s.logger.Error(err.Error())
+		return nil, err
+	}
+
+	if _, ok := cm.Data[DEFAULT_SCHEMA]; !ok {
+		return nil, fmt.Errorf("default schema %s missing", DEFAULT_SCHEMA)
+	}
```
nsklikas, line 266: I still don't understand why this is needed. If the default schema is missing, that means that the config is wrong. We would want to fix the broken config rather than throw an error.

pkg/schemas/service.go
```diff
@@ -223,6 +227,59 @@ func (s *Service) DeleteSchema(ctx context.Context, ID string) error {
...
+
+	if schema.ID == DEFAULT_SCHEMA {
+		return nil, fmt.Errorf("schema with ID %s not available", schema.ID)
+	} else if _, ok := cm.Data[schema.ID]; !ok && schema.ID != cm.Data[DEFAULT_SCHEMA] {
```
nsklikas, line 270: If the user provides the same schema ID, why would we throw an error? The second check should be removed.

I think that the logic here should be just that the `schema.ID` provided is in the map and that it does not have the `DEFAULT_SCHEMA` value.

pkg/schemas/service.go
nsklikas: I don't think we should care if the value of the `default.schema` changes, it makes no difference. 
Being idempotent means that even if the user performs the same action many times, the result will be the same. With the current logic, if the user updates the default schema they will get a `200` response, but if they try to make the same call again they will get a `500`.

bencekov: I see your point. I modified the code.

pkg/schemas/service.go
bencekov: For me it wasn't obvious that a bad config shouldn't throw an error, but I accept your reasoning. fixed.

pkg/schemas/service.go
bencekov: This was misguided. I fixed it.

nsklikas (approved):

PR #185, opened 2024-01-22
Introduce cobra CLI and basic authorization functionality

nsklikas:
IAM-657

- Migrate to cobra CLI
- Introduce openfga module
- Introduce authorization module
- Add tracing interface


cmd/createFgaModel.go
```diff
@@ -0,0 +1,63 @@
...
+	createFgaModelCmd.PersistentFlags().String("fga-api-url", "", "The openfga API URL")
+	createFgaModelCmd.PersistentFlags().String("fga-api-token", "", "The openfga API token")
+	createFgaModelCmd.PersistentFlags().String("fga-store-id", "", "The openfga store to create the model in")
+	createFgaModelCmd.MarkPersistentFlagRequired("fga-api-url")
+	createFgaModelCmd.MarkPersistentFlagRequired("fga-api-token")
+	createFgaModelCmd.MarkPersistentFlagRequired("fga-store-id")
+}
+
+func createModel(apiUrl, apiToken, storeId string) {
+	logger := logging.NewNoopLogger()
```
shipperizer, line 41: remind me again, why don't we want the normal logger?

internal/authorization/auth_model.go
```diff
@@ -0,0 +1,5 @@
...
...mber","type":"group"}]},"can_create":{},"can_delete":{"directly_related_user_types":[{"type":"user"},{"relation":"assignee","type":"role"},{"relation":"member","type":"group"}]},"can_edit":{},"can_list":{},"can_view":{},"editor":{"directly_related_user_types":[{"type":"user"},{"relation":"assignee","type":"role"},{"relation":"member","type":"group"}]},"viewer":{"directly_related_user_types":[{"type":"user"},{"relation":"assignee","type":"role"},{"relation":"member","type":"group"}]}}},"relations":{"admin":{"this":{}},"can_create":{"union":{"child":[{"computedUserset":{"relation":"editor"}},{"computedUserset":{"relation":"admin"}}]}},"can_delete":{"union":{"child":[{"this":{}},{"computedUserset":{"relation":"admin"}}]}},"can_edit":{"union":{"child":[{"computedUserset":{"relation":"editor"}},{"computedUserset":{"relation":"admin"}}]}},"can_list":{"union":{"child":[{"computedUserset":{"relation":"viewer"}},{"computedUserset":{"relation":"admin"}}]}},"can_view":{"union":{"child":[{"computedUserset":{"relation":"viewer"}},{"computedUserset":{"relation":"admin"}}]}},"editor":{"this":{}},"viewer":{"union":{"child":[{"this":{}},{"computedUserset":{"relation":"editor"}}]}}},"type":"api"}]}`
```
shipperizer, line 5: have a look at making it a yaml if possible

internal/authorization/authorization.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/canonical/identity-platform-admin-ui/internal/logging"
+	"github.com/canonical/identity-platform-admin-ui/internal/monitoring"
+	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
+	fga "github.com/openfga/go-sdk"
+)
+
+var ErrInvalidAuthModel = fmt.Errorf("Invalid authorization model schema")
+
+type Authorizer struct {
+	Client AuthzClientInterface
```
shipperizer, line 17:
```suggestion
	client AuthzClientInterface
```

shipperizer: it can be a private variable

internal/authorization/auth_model.go
shipperizer: ah I saw it in the other file, wondering if it s better to embed it and then call the function at bootstrap

structure-tests.yaml
```diff
@@ -32,5 +32,5 @@ fileExistenceTests:
   shouldExist: true
 commandTests:
   - name: "application version"
-    command: "/usr/bin/identity-platform-admin-ui"
-    args: ["-version"]
+    command: "/usr/bin/identity-platform-admin-ui version"
+    # args: ["version"]
```
shipperizer, line 36: I was expecting to use args....interesting it works without

cmd/createFgaModel.go
nsklikas: When I run an one-off cli command I don't expect to see the timestamp and all the extra info the logger provides. 

I can change it if you want, just felt counterintuitive.

internal/authorization/auth_model.go
nsklikas: We use the fga CLI to transform it, we need something like this https://github.com/openfga/cli/blob/main/internal/authorizationmodel/model.go#L156. We can either install that module or copy/paste it, neither look very pretty.

cmd/createFgaModel.go
shipperizer: this would be called by a juju action or init container, still valuable to get the logs into loki imho

structure-tests.yaml
nsklikas: it was leftover from debugging, removed

cmd/createFgaModel.go
nsklikas: This would mean that you would configure this write the command output to the file that loki reads from or that you would configure loki to read the stdout form the init container, so I would need to parse the env vars or add a config option for piping the output to a log file.

shipperizer: as discussed, leave it be, we can always add it later

internal/authorization/auth_model.go
```diff
@@ -0,0 +1,35 @@
...
+	if err != nil {
+		panic(fmt.Errorf("failed to transform due to %w", err))
+	}
+
+	err = json.Unmarshal(bytes, &jsonAuthModel)
+	if err != nil {
+		panic(fmt.Errorf("failed to transform due to %w", err))
+	}
+	return jsonAuthModel
+}()
```
shipperizer, line 35: due to how `embed` works (which is not immediate when dealing with paths imho) , let's bring up the whole setup and make sure it works

internal/logging/noop.go
```diff
@@ -0,0 +1,9 @@
+package logging
+
+import (
+	"go.uber.org/zap"
+)
+
+func NewNoopLogger() *zap.SugaredLogger {
+	return zap.NewNop().Sugar()
+}
```
shipperizer, line 9: i'd say overkill but happy to keep it

deployments/kubectl/deployment.yaml
```diff
@@ -44,5 +44,25 @@ spec:
...
+          - "42"
+          - --name
+          - admin-ui
+          volumeMounts:
+          - name: workdir
+            mountPath: "/fga-store-info"
+      dnsPolicy: Default
+      volumes:
+      - name: workdir
+        emptyDir: {}
```
shipperizer, line 66: will we still need this or we go full manual with portforwarding, `fga store create` and `kubect edit cm`?

internal/authorization/auth_model.go
nsklikas: I have tested this

deployments/kubectl/deployment.yaml
nsklikas: removed

nsklikas: removed it

shipperizer (approved):

PR #187, opened 2024-01-25
Add command for creation of admin users

nsklikas:
IAM-656

Add the command for creation and deletion of admin users. Whenever a new role is created we will need to create a tuple relating it with `privileged:user`
image


cmd/create.go
```diff
@@ -12,20 +19,53 @@ var createCmd = &cobra.Command{
...
-	// and all subcommands, e.g.:
-	// createCmd.PersistentFlags().String("foo", "", "A help for foo")
+	if user == "" {
+		user = randUserName()
+	}
+
+	err = auth.CreateAdmin(context.Background(), user)
+	if err != nil {
+		panic(fmt.Errorf("failed to create user: %w", err))
+	}
```
nsklikas, line 64: If the user already exists the user will get an error like this 
image

Should we handle this more gracefully?

cmd/create.go
```diff
@@ -12,20 +19,54 @@ var createCmd = &cobra.Command{
...
+	cfg := openfga.NewConfig(scheme, host, storeId, apiToken, "", false, tracer, monitor, logger)
+	fgaClient := openfga.NewClient(cfg)
+	auth := authorization.NewAuthorizer(fgaClient, tracer, monitor, logger)
 
-	// Cobra supports Persistent Flags which will work for this command
-	// and all subcommands, e.g.:
-	// createCmd.PersistentFlags().String("foo", "", "A help for foo")
+	if user == "" {
+		user = randUserName()
```
shipperizer, line 59: I'd keep this inline, no need for extra func

cmd/create.go
shipperizer: what about logging the error and the os.exit?

cmd/remove.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		panic(err)
+	}
+	cfg := openfga.NewConfig(scheme, host, storeId, apiToken, "", false, tracer, monitor, logger)
+	fgaClient := openfga.NewClient(cfg)
+	auth := authorization.NewAuthorizer(fgaClient, tracer, monitor, logger)
+
+	err = auth.RemoveAdmin(context.Background(), user)
+	if err != nil {
+		panic(fmt.Errorf("failed to remove admin: %w", err))
```
shipperizer, line 62: same here, log the error then os.exit

internal/authorization/admin.go
```diff
@@ -0,0 +1,34 @@
...
+	err := a.client.WriteTuple(ctx, user, "admin", "privileged:user")
+	return err
+}
+
+func (a *Authorizer) RemoveAdmin(ctx context.Context, username string) error {
+	ctx, span := a.tracer.Start(ctx, "authorization.Authorizer.RemoveAdmin")
+	defer span.End()
+
+	user := fmt.Sprintf("user:%s", username)
+	err := a.client.DeleteTuple(ctx, user, "admin", "privileged:user")
```
shipperizer, line 22: let's make the `privileged:user` a constant

cmd/create.go
nsklikas: I think I should remove this, not sure if it makes sense to auto-generate a user ID.

internal/authorization/admin.go
```diff
@@ -0,0 +1,36 @@
...
+
+func (a *Authorizer) CheckAdmin(ctx context.Context, username string) (bool, error) {
+	ctx, span := a.tracer.Start(ctx, "authorization.Authorizer.CheckAdmin")
+	defer span.End()
+
+	user := fmt.Sprintf("user:%s", username)
+	allowed, err := a.Check(ctx, user, "admin", ADMIN_OBJECT)
+
+	return allowed, err
+}
```
shipperizer, line 36: this is a bit confusing, as `struct` is defined in a different file

you could be creating a separate struct for these functions and use [composition](https://www.codecademy.com/resources/docs/go/composition)

that would make it siloed

nsklikas: makes sense, changed it.

shipperizer (dismissed):

shipperizer (approved):

PR #189, opened 2024-02-01
IAM 655 - roles API

shipperizer:
IAM-655: roles API following [openapi spec v0.0.8 ](https://github.com/canonical/openfga-admin-openapi-spec/pull/9/files)

- feat: upgrade openfga model c49abd5
- feat: handlers for roles API 114b284
- feat: create token pagination extractor 215b6cb
- feat: create roles service c796135
- feat: add Urn type f7d33e2
- feat: wire up roles API in web application 16ba352
- build: update dependencies b4da42d
- fix: use sync.Map for race conditions 603a7e1
- fix: switch to use WriteTuples instead of WriteTuple ba8a624
- docs: add brief description in public func and methods 584f406

** MISSING ** 

* handling of loose goroutines on `DELETE /roles` and `GET /roles/<id>/entitlements` calls


pkg/roles/handlers.go
```diff
@@ -0,0 +1,458 @@
...
+		types.Response{
+			Message: fmt.Sprintf("Created role %s", role.ID),
+			Status:  http.StatusCreated,
+		},
+	)
+}
+
+// handleUpdate is not implemented by choice, product might decide to do it to enhcance
+// role metadata, we do not support anything on top of simple ID attribute and this is
+// not changeable right now due to coupled implemntation with OpenFGA
```
alesstimec, line 194: s/implemntation/implementation/

alesstimec: if this is not implemented, then just omit this method.. or return a not implemented error

pkg/roles/handlers.go
```diff
@@ -0,0 +1,458 @@
...
+}
+
+// TODO @shipperizer we need to remove all relationships leading to the role
+func (a *API) handleRemove(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+
+	ID := chi.URLParam(r, "id")
+
+	err := a.service.DeleteRole(r.Context(), ID)
+
```
alesstimec, line 216: dd

pkg/roles/service.go
```diff
@@ -0,0 +1,294 @@
...
+
+package roles
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/canonical/identity-platform-admin-ui/internal/authorization"
```
alesstimec, line 12: imports need to be sorted

pkg/web/router.go
```diff
@@ -6,10 +6,12 @@ package web
 import (
 	"net/http"
 
+	"github.com/canonical/identity-platform-admin-ui/internal/authorization"
```
alesstimec, line 9: imports need to be sorted

alesstimec (changes requested):
this PR tries to do too much in one go. i suggest we break it up into more manageable chunks..

shipperizer:
> this PR tries to do too much in one go. i suggest we break it up into more manageable chunks.. 

splitting all the stuff in the internal/openfga should free up some changes

at the same time, anything inside cmd/dummy is not going to be committed, see WIP commit

cmd/dummy/main.go
```diff
@@ -0,0 +1,160 @@
...
+	w.WriteHeader(http.StatusOK)
+	json.NewEncoder(w).Encode(
+		types.Response{
+			Message: fmt.Sprintf("requested %s", r.URL.Path),
+			Status:  http.StatusOK,
+		},
+	)
+}
+
+func RegisterEndpoints(mux *chi.Mux) {
```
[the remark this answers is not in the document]
shipperizer: anything in dummy is simply for dev testing purposes, it a inside the WIP commit and should be dropped when PR is ready to be reviewed

cmd/dummy/main.go
```diff
@@ -0,0 +1,160 @@
+package main
```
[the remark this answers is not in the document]
shipperizer: this won't go in, part of WIP commit

internal/authorization/middlewares.go
```diff
@@ -0,0 +1,205 @@
...
+	SchemeConverter
+	RoleConverter
+
+	monitor monitoring.MonitorInterface
+	logger  logging.LoggerInterface
+}
+
+func (mdw *Middleware) transformToken(token string) *User {
+	user := new(User)
+	user.ID = "admin"
```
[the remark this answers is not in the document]
shipperizer: unfortunately authorization comes in a different tranche of work than this one, that's where the proper implementation will happen

for the time being we need a way to check users and this was the quickest way to make it happen, this should not go in the final PR setup but at the same time we needed to get the dummy application up and running fit the poc

internal/authorization/middlewares.go
```diff
@@ -0,0 +1,205 @@
...
+)
+
+type UserContext string
+type AdminContext string
+
+type User struct {
+	ID string
+}
+
+// Middleware is the monitoring middleware object implementing Prometheus monitoring
```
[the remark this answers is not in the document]
shipperizer: bad copy/paste

pkg/roles/service.go
```diff
@@ -0,0 +1,310 @@
...
+		// if there are more pages, keep going with the loop
+		if cToken = r.ContinuationToken; cToken != "" {
+			continue
+		}
+
+		// TODO @shipperizer understand if better breaking at every cycle or reverting if clause
+		break
+	}
+
+	if err := s.ofga.DeleteTuples(ctx, permissions...); err != nil {
```
nsklikas, line 296: Perhaps it is an unrealistic case, but I am a little worried that this may error if too many permissions are returned. I mean that openfga chooses not to return the permissions all at once, it is possible that it will not be able to handle them all at once either. In the API docs there is no mention of a limit, but I think that we should test it. 

Disregard this if it does not make sense or leave a TODO to look into it in the future. 

If this could cause an issue we could create a go routine for deleting the permissions for every continuation token we get, eg:
```
for {
  // Get the permissions using the continuation token
  permissions, error := getPermissions(continuationToken)
  // Delete permissions without blocking
  go deletePermissions(permissions)
  cToken = permissions.ContinuationToken
  if cToken == "" {
    break
  }
  // wait for all routines to return
}
```

nsklikas (approved):
I did not test it, but LGTM overall other than some minor comments

pkg/roles/service.go
```diff
@@ -0,0 +1,310 @@
...
+}
+
+func (s *Service) ListPermissions(ctx context.Context, ID string, continuationTokens map[string]string) ([]string, map[string]string, error) {
+	ctx, span := s.tracer.Start(ctx, "roles.Service.ListPermissions")
+	defer span.End()
+
+	permissionsMap := sync.Map{}
+	tokensMap := sync.Map{}
+
+	permTypes := []string{"role", "group", "identity", "scheme", "provider", "client"}
```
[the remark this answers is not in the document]
shipperizer: can't be a constant unfortunately as it s a slice, I can put it as a service method

pkg/roles/service.go
```diff
@@ -0,0 +1,310 @@
...
+
+	return permissions, r.GetContinuationToken(), nil
+}
+
+func (s *Service) removePermissionsByType(ctx context.Context, ID, pType string) {
+	ctx, span := s.tracer.Start(ctx, "roles.Service.removePermissionsByType")
+	defer span.End()
+
+	cToken := ""
+	assigneeRelation := fmt.Sprintf("role:%s#%s", ID, ASSIGNEE_RELATION)
```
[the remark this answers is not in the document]
shipperizer: that s fair, let me add an issue for it

pkg/roles/service.go
shipperizer: issue is we have already enough loose goroutines per request (6) so this would be really unmanageable with the current implementation

we need to find a better setup to deal with the request load, that would be the first step

ideally we would create a `TaskManager` object that deals with a pool of goroutines and allows for this background work to happen, but i can see it quite complicated 

i'll definitely add a TODO

pkg/roles/service.go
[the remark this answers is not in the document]
shipperizer: offloaded to a method `s.permissionTypes()`

pkg/roles/service.go
shipperizer: https://github.com/canonical/identity-platform-admin-ui/issues/217

nsklikas (approved):

BarcoMasile (approved):

PR #193, opened 2024-02-07
feat: passing openfga store and model id to admin service

wood-push-melon:
This pull request tries to solve the issue in the provisioning development setup. The admin service needs the openFGA store ID and model ID to be passed in.

The proposed setup uses a k8s job, instead of using init containers in Admin UI deployment, to create openFGA store and model and share them with the admin service by using persistent volume. This is specifically to avoid potential race conditions that could be caused by multiple Admin UI replicas.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,12 +1,33 @@
 #!/bin/sh
 
-set -x
-# script requires the availability of rockcraft, skopeo, yq and docker in the host system
-# it also requires sudo permissions to run skopeo
+# The script requires:
+# - rockcraft
+# - skopeo with sudo privilege
+# - yq
+# - docker
 
-# export version=$(yq -r '.version' rockcraft.yaml)
+set -e
+
+rockcraft="rockcraft.yaml"
+rockcraft_backup="rockcraft_bak.yaml"
+
+restore() {
+    mv "$rockcraft_backup" "$rockcraft"
+    rm -f "$rockcraft_backup"
+}
+trap 'restore' INT TERM EXIT
+
+# The ROCK image needs certain utilities to
+# - create OpenFGA store and authorization model
+# - export OpenFGA store ID and authorization model ID to Admin UI service
+cp "$rockcraft" "$rockcraft_backup"
+yq -i \
+  '.base = "ubuntu@22.04", .parts |= ({"utils": {"plugin": "nil", "stage-packages": ["curl", "jq"]}} + .)' \
+  "$rockcraft"
 rockcraft pack -v
 
-sudo skopeo --insecure-policy copy "oci-archive:identity-platform-admin-ui_$(yq -r '.version' rockcraft.yaml)_amd64.rock" docker-daemon:$IMAGE
+sudo skopeo --insecure-policy \
+  copy "oci-archive:identity-platform-admin-ui_$(yq -r '.version' rockcraft.yaml)_amd64.rock" \
+  docker-daemon:"$IMAGE"
 
-docker push $IMAGE
+docker push "$IMAGE"
--- a/deployments/kubectl/configMap.yaml
+++ b/deployments/kubectl/configMap.yaml
@@ -112,3 +112,33 @@ data:
                 "type": "object"
             }
         }
+
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: openfga-init
+data:
+  init.sh: |
+    #!/bin/sh
+
+    set -e
+
+    OPENFGA_URL='http://openfga.default.svc.cluster.local:8080'
+
+    until [ "$(curl -s -o /dev/null -w "%{http_code}" $OPENFGA_URL/stores)" -eq 200 ]; do
+      sleep 5
+    done
+
+    curl -sS -X POST ${OPENFGA_URL}/stores \
+      -H 'Content-Type: application/json' \
+      -d '{"name": "dev"}' | jq -r '.id' > /data/OPENFGA_STORE_ID
+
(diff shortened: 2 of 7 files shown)


deployments/kubectl/configMap.yaml
```diff
@@ -112,3 +112,286 @@ data:
...
+      sleep 5
+    done
+
+    curl -sS -X POST ${OPENFGA_URL}/stores \
+      -H 'Content-Type: application/json' \
+      -d '{"name": "dev"}' | jq -r '.id' > /data/OPENFGA_STORE_ID
+
+    curl -sS -X POST ${OPENFGA_URL}/stores/"$(cat /data/OPENFGA_STORE_ID)"/authorization-models \
+      -H 'Content-Type: application/json' \
+      -d  @/schemas/schema.json | jq -r '.authorization_model_id' > /data/OPENFGA_AUTHORIZATION_MODEL_ID
```
nsklikas, line 397: Wouldn't it be simpler if we set this as the command in the `openfga-init` job? Something like
```
sh -c 'set -e
OPENFGA_URL='http://localhost:8080'
until [ "$(curl -s -o /dev/null -w "%{http_code}" $OPENFGA_URL/stores)" -eq 200 ]; do
  sleep 5
done
curl -sS -X POST ${OPENFGA_URL}/stores \
  -H 'Content-Type: application/json' \
  -d '{"name": "dev"}' | jq -r '.id' > /data/OPENFGA_STORE_ID
curl -sS -X POST ${OPENFGA_URL}/stores/"$(cat /data/OPENFGA_STORE_ID)"/authorization-models \
  -H 'Content-Type: application/json' \
  -d  @/schemas/schema.json | jq -r '.authorization_model_id' > /data/OPENFGA_AUTHORIZATION_MODEL_ID'
```

Or would it pollute the file too much?

deployments/kubectl/configMap.yaml
```diff
@@ -112,3 +112,286 @@ data:
...
             }
         }
+
+---
+apiVersion: v1
+kind: ConfigMap
+metadata:
+  name: openfga-schema
+data:
+  schema.json: |
```
nsklikas, line 122: I don't like this, we would have to update this json every time we make a change to the schema. Can't we use the CLI: `identity-platform-admin-ui create-fga-model` for this? (we could check if the `OPENFGA_AUTHORIZATION_MODEL_ID` file exists and if not we run this command)

build.sh
```diff
@@ -1,12 +1,28 @@
...
+rockcraft_backup="rockcraft_bak.yaml"
+
+restore() {
+    mv "$rockcraft_backup" "$rockcraft"
+    rm -f "$rockcraft_backup"
+}
+trap 'restore' INT TERM EXIT
+
+cp "$rockcraft" "$rockcraft_backup"
+sed -i 's/^base: .*/base: ubuntu@22.04/' "$rockcraft"
```
nsklikas, line 21: For future reference can you add a comment here saying why we need to do this (something like `use ubuntu base image because we need a shell to export the env vars from a file`)?

deployments/kubectl/deployment.yaml
```diff
@@ -19,9 +19,20 @@ spec:
         prometheus.io/scrape: "true"
         prometheus.io/port: "8000"
     spec:
+      initContainers:
+        - name: init
+          image: busybox
+          command: [ "/bin/sh", "-c", "while [ ! -s /data/openfga/OPENFGA_STORE_ID ] || [ ! -s /data/openfga/OPENFGA_AUTHORIZATION_MODEL_ID ]; do sleep 5; done" ]
+          volumeMounts:
+            - mountPath: /data/openfga
+              name: openfga-pv-data
```
nsklikas, line 28: Alternatively, could we use an image with kubectl an initContainer, to update the deployment and add the `OPENFGA_STORE_ID` and `OPENFGA_AUTHORIZATION_MODEL_ID` as env vars to the `identity-platform-admin-ui` container?

nsklikas (commented):
Left a couple of comments. My main problem is with pasting the schema in a configMap. Let's discuss this further before/after the daily

deployments/kubectl/configMap.yaml
wood-push-melon: I think it's too much if we move the script to the `commands`. Generally, it is a common way to use ConfigMap to share the local script with remote Pod.

deployments/kubectl/deployment.yaml
wood-push-melon: I intentionally avoid using `kubectl` inside any Pod to modify k8s resources on the fly because I think it's not a recommended way to do things with k8s. We can discuss it.

deployments/kubectl/configMap.yaml
```diff
@@ -112,3 +112,33 @@ data:
...
+      -H 'Content-Type: application/json' \
+      -d '{"name": "dev"}' | jq -r '.id' > /data/OPENFGA_STORE_ID
+
+    openfga_store_id=$(cat /data/OPENFGA_STORE_ID)
+    /usr/bin/identity-platform-admin-ui create-fga-model \
+      --fga-api-url "$OPENFGA_URL" \
+      --fga-api-token "42" \
+      --fga-store-id "$openfga_store_id"
+
+    curl -sS ${OPENFGA_URL}/stores/"$openfga_store_id"/authorization-models \
```
wood-push-melon, line 143: I `curl` the OpenFGA endpoint to get the authorization model created by the CLI command above instead of parsing the CLI command output, in order to avoid future changes in the command line which could introduce the instability of the script.

nsklikas (approved):
lgtm

PR #194, opened 2024-02-16
(ReBAC library) Initialize ReBAC Admin UI Backend library

babakks:
This PR creates a separate Go module encompassing the *ReBAC Admin UI Backend` library. Essentially, this library implements HTTP handlers according to the ReBAC Admin UI OpenAPI spec and delegates the actual operation (authorization or manipulating OpenFGA tuples) to custom implementations plugged in via pre-defined interfaces.

Since the library is meant to be a reusable component and will later be extracted into a separate repository, we try to keep it as a separate/independent Go module. Note that, to make it play nice with the main Go module in the repo's root, we had to use Go Workspaces. After the library is extracted, we can delete the Go Workspace.

Fixes CSS-7248


shipperizer:
@babakks can u rewrite the commits to follow the conventional commits standard?

reason is that our release process uses those for drafting a release

rebac-admin-backend/script/pull-spec.sh
```diff
@@ -0,0 +1,26 @@
+#!/usr/bin/env sh
```
shipperizer, line 1: why not using a git submodule?

rebac-admin-backend/api/openapi.yaml
```diff
@@ -0,0 +1,1607 @@
+openapi: 3.1.0
```
alesstimec, line 1: does this need to be part of this spec as it already lives in its own repo?

rebac-admin-backend/script/pull-spec.sh
alesstimec: +1

rebac-admin-backend/internal/api/v1/types.go
```diff
@@ -0,0 +1,390 @@
...
+// CapabilitiesResponse defines model for CapabilitiesResponse.
+type CapabilitiesResponse struct {
+	Data []struct {
+		Endpoint string                            `json:"endpoint"`
+		Methods  []CapabilitiesResponseDataMethods `json:"methods"`
+	} `json:"data"`
+}
+
+// CapabilitiesResponseDataMethods defines model for CapabilitiesResponse.Data.Methods.
+type CapabilitiesResponseDataMethods string
```
alesstimec, line 39: `type CapabilitiesResponseDataMethod string`

alesstimec: ah.. it's generated..

babakks:
> @babakks can u rewrite the commits to follow the conventional commits standard?
> 
> reason is that our release process uses those for drafting a release

Yeah, sure. I usually try keeping my commits as small as possible, so at times it might be a bit difficult to pick/find an accurate commit message type (e.g., `fix` or `feat`). But I'll do my best.

shipperizer:
> > @babakks can u rewrite the commits to follow the conventional commits standard?
> > 
> > reason is that our release process uses those for drafting a release
> 
> Yeah, sure. I usually try keeping my commits as small as possible, so at times it might be a bit difficult to pick/find an accurate commit message type (e.g., `fix` or `feat`). But I'll do my best.

don't sweat it, just put fix if it s a minor change or chore if u think it s irrelevant

rebac-admin-backend/script/pull-spec.sh
babakks: That was my first approach, but then I noticed I just need the `openapi.yaml` file and not the rest of the repository.

But, now as I see, we can just drop this copy of the spec and embed it in the auto-generated code (See the `scope` option [here](https://github.com/deepmap/oapi-codegen#using-oapi-codegen)). This way we can even serve that spec, which could be useful for debugging/inspection to see what's is the exact spec that a backend implementers.

@alesstimec @shipperizer Sounds good?

babakks:
@alesstimec @shipperizer Sorry for the `--force` push.

babakks:
I removed the OpenAPI spec copy, and now we have it embedded within the auto-generated types.

babakks:
I also removed the Go Workspace becasuse it causes some conflicts with the other parts of the
respository (e.g., go generate directives with -mod=mod option). Since we don't want to touch other parts of the repo for this library, we'd better off keeping the library as a package inside the main Go module.

rebac-admin-backend/script/pull-spec.sh
shipperizer: ok, this is probably shooting for the starts but i ll drop it anyway

ideally this gets updated automatically at every given update of the underlying source (so the openapi-repo), updated as a GH action runs and creates a PR

codebase is supposed ot be only stable when it s relaeased, so we wouldn't incur in any issue by having this kind of machinery (which is not needed as part of this PR, not asking for that dont worry)

i wouldnt be that bothered about having  a full repo as a submodule even if we used only one file, reason is we might have to offload the makefile to that repo  and allow different applications to call the codegen command 

anyhow, as a starting solution i m more than happy, would the `spec` bit work off a raw.github url?

babakks: > anyhow, as a starting solution i m more than happy, would the spec bit work off a raw.github URL?

Yeah, something like that. If you look at the generated file, it's the `gzip` (base64 encoded). I don't implement the endpoint to serve the spec just now. If we want to do it, we need to define it in the ReBAC admin spec as well.

babakks:
Just one more update. I also generated the server part (handlers and routes). I think it's done now and we can merge it if it's okay with you. @alesstimec @shipperizer

rebac-admin-backend/script/pull-spec.sh
babakks: https://github.com/canonical/identity-platform-admin-ui/blob/81c8459c86f6f1aebb19d4b38d7662e368955fcd/rebac-admin-backend/internal/api/v1/types.go#L401

babakks:
**UPDATE**: The `/swagger.json` endpoint is now added to the ReBAC Admin OpenAPI spec.

shipperizer (commented):
@babakks dont hate me mate, i m a bit nitpicky with commits

from what i see there are 3 commits we can group our changes on

* machinery for pulling and generating code (`chore: ` or `build: `)
* generated code (`feat: `)
* dependencies update (`chore: ` or `build:`

do u mind rewriting the commits?

then it can go in imo

babakks:
Reworded the commits. Please have a look. @shipperizer

shipperizer:
@babakks I think u forgot to force push, there are still 31 commits instead of 3

babakks:
Sorry, I thought your suggesting commit message rewording. I'll rebase and push with 3 or so commits.

babakks:
@shipperizer Done.

shipperizer (approved):

PR #196, opened 2024-02-19
(ReBAC library) Add main types

babakks:
This PR adds the main types of the library (e.g., params, constructor, some abstractions) and also creates HTTP handlers. A few tests are added to verify it works fine with both Golang standard HTTP multiplexer and Chi Mux.

This PR is based on the other branch (pushed in #194). That's why you see a lot of changes, but once the other PR lands, this one will shrink.

Fixes CSS-7249

### DIFF
--- a/rebac-admin-backend/v1/core.go
+++ b/rebac-admin-backend/v1/core.go
@@ -1,0 +1,43 @@
+// Copyright 2024 Canonical Ltd.
+
+// Package v1 provides HTTP handlers that implement the ReBAC Admin OpenAPI spec.
+// This package delegates authorization and data manipulation to user-defined
+// "backend"s that implement the designated abstractions.
+package v1
+
+import (
+	"net/http"
+	"strings"
+
+	"github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
+)
+
+// ReBACAdminBackendParams contains references to user-defined implementation
+// of required abstractions, called "backend"s.
+type ReBACAdminBackendParams struct {
+	GroupsService       GroupsServiceBackend
+	GroupsAuthorization GroupsAuthorizationBackend
+}
+
+// ReBACAdminBackend represents the ReBAC admin backend as a whole package.
+type ReBACAdminBackend struct {
+	params  ReBACAdminBackendParams
+	service resources.ServerInterface
+}
+
+// NewReBACAdminBackend returns a new ReBACAdminBackend instance, configured
+// with given backends.
+func NewReBACAdminBackend(params ReBACAdminBackendParams) *ReBACAdminBackend {
+	return &ReBACAdminBackend{
+		params:  params,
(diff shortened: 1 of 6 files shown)


rebac-admin-backend/v1/core.go
```diff
@@ -0,0 +1,43 @@
...
+type ReBACAdminBackendParams struct {
+	GroupsService       GroupsServiceBackend
+	GroupsAuthorization GroupsAuthorizationBackend
+}
+
+// ReBACAdminBackend represents the ReBAC admin backend as a whole package.
+type ReBACAdminBackend struct {
+	params  ReBACAdminBackendParams
+	service *service
+}
```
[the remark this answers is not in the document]
babakks: No, it's not auto-generated, and it's not because of the embedded struct. That was out of habit, mostly because you can't go wrong with pointers (e.g., there'll be no unexpected struct copies). But in this case we can use a non-pointer field. I'll update the code.

babakks:
Changed the pointer field to value type, and rebased on top of my other branch. @shipperizer

rebac-admin-backend/v1/core.go
```diff
@@ -0,0 +1,43 @@
...
+// of required abstractions, called "backend"s.
+type ReBACAdminBackendParams struct {
+	GroupsService       GroupsServiceBackend
+	GroupsAuthorization GroupsAuthorizationBackend
+}
+
+// ReBACAdminBackend represents the ReBAC admin backend as a whole package.
+type ReBACAdminBackend struct {
+	params  ReBACAdminBackendParams
+	service service
```
shipperizer, line 25: like this it will be impossible to mock
I don't have the big picture but I guess this holds all the logic, if that's the case we need that as interface

babakks: We don't need to mock this object (explained below), but, nevertheless, liked the idea in general. This `service` field already implements an auto-generated interface called `ServerInterface`, so we can just replace the type with that. 

About mocking `service`, note that it's the outmost layer that receives the requests (dispatched by the HTTP mux) and therefore we wouldn't normally want to mock it. Unless the code needs to invoke other endpoints in a loopback way (I mean not by invoking internal underlying methods, but going through the HTTP mux), which is not imaginable here.

babakks: `ServerInterface` is like this:

```go
// ServerInterface represents all server handlers.
type ServerInterface interface {
	// List configured authentication providers.
	// (GET /authentication)
	GetAuthentication(w http.ResponseWriter, r *http.Request, params GetAuthenticationParams)
	// Configure a new authentication provider.
	// (POST /authentication)
	PostAuthentication(w http.ResponseWriter, r *http.Request)
	// Returns the list of supported identity providers.
	// (GET /authentication/providers)
	// ...
}
```

PR #197, opened 2024-02-20
(ReBAC library) Implement proper error formatting

babakks:
This PR applies proper error formatting to match the OpenAPI spec, by implementing a custom `ErrorHandlerFunc` function. This also introduces some generic error types (e.g., `UnauthorizedError` or `NotFoundError`) which should be used by custom backends).

Some test cases are added to verify errors added are correctly formatted to a `Response` type.

Fixes CSS-7327

### DIFF
--- a/rebac-admin-backend/v1/error.go
+++ b/rebac-admin-backend/v1/error.go
@@ -1,0 +1,17 @@
+// Copyright 2024 Canonical Ltd.
+
+package v1
+
+// UnauthorizedError represents unauthorized access error.
+type UnauthorizedError struct{}
+
+func (e *UnauthorizedError) Error() string {
+	return "unauthorized"
+}
+
+// NotFoundError represents missing entity error.
+type NotFoundError struct{}
+
+func (e *NotFoundError) Error() string {
+	return "not found"
+}
--- a/rebac-admin-backend/v1/response.go
+++ b/rebac-admin-backend/v1/response.go
@@ -1,0 +1,71 @@
+// Copyright 2024 Canonical Ltd.
+
+package v1
+
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
+)
+
+// writeErrorResponse writes the given err in the response with format defined
+// by the OpenAPI spec.
+func writeErrorResponse(w http.ResponseWriter, r *http.Request, err error) {
+	resp := getErrorResponse(err)
+
+	body, err := json.Marshal(resp)
+	if err != nil {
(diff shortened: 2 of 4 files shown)


rebac-admin-backend/v1/response_test.go
```diff
@@ -0,0 +1,87 @@
...
+
+package v1
+
+import (
+	"errors"
+	"net/http"
+	"testing"
+
+	"github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
+	qt "github.com/frankban/quicktest"
```
shipperizer, line 11: i understand that this will be a repo on its own eventually, but for now can we stick with `uber/gomock` and std library for testing?

alesstimec: -1 on using the std library.. quicktest serves use very well in all of our projects AND we know the author :)

alesstimec: but imports still need sorting

rebac-admin-backend/v1/error.go
```diff
@@ -0,0 +1,17 @@
+// Copyright 2024 Canonical Ltd.
+
+package v1
+
+// UnauthorizedError represents unauthorized access error.
+type UnauthorizedError struct{}
```
alesstimec, line 6: suggestion
```
var (
   UnauthorizedError = errors.New("unauthorized")
   NotFoundError = errors.New("not found")
)
```

rebac-admin-backend/v1/response.go
```diff
@@ -0,0 +1,71 @@
...
+		w.WriteHeader(http.StatusInternalServerError)
+		w.Write([]byte("unexpected marshalling error"))
+	}
+
+	w.WriteHeader(int(resp.Status))
+	w.Write(body)
+}
+
+// getErrorResponse returns a Response instance filled with the given error.
+func getErrorResponse(err error) resources.Response {
```
alesstimec, line 28: nitpick/suggestion: mapErrorResponse

alesstimec (approved):
LGTM modulo a few comments

rebac-admin-backend/v1/error.go
babakks: I thought of that, but we might want to add more information to the errors down the road. For example, in case of `NotFoundError`, for some endpoints there are more than one entity that could be missing. For example, for `PATCH /groups/{id}/identities`, the group ID or any of the identity IDs could be missing. So, for more descriptive errors we might want to have such information attached.

I haven't yet added fields to the error types, because we need to figure it out yet; either we define some fields on them, or we only change the string message. In any way, we'd need a separate error struct type.

BarcoMasile (approved):

PR #200, opened 2024-02-26
feat: add `/identities` API set handlers implementation

BarcoMasile:
## Description
First implementation of the `/identities` API set handlers.
This PR leverages the work done on the most recent PRs on the OpenAPI spec repo ([PR 20](https://github.com/canonical/openfga-admin-openapi-spec/pull/20) + [PR 21](https://github.com/canonical/openfga-admin-openapi-spec/pull/21) + [PR 22](https://github.com/canonical/openfga-admin-openapi-spec/pull/22)), spawned from issues arisen while working on this PR.

This PR presents many microcommits for two reasons:
- ease of review
- group logically bound edits

### Important
- being this the first PR in a series for the Endpoints work, many needed constructs/functions are part of this work, increasing the number of added/changed lines
- of all the endpoints, **three are 3 still missing due to some issues in the current OpenAPi spec**, that could not be addressed in the middle of the work for this PR and will be addressed shortly in a new PR:
  - `PATCH /identities/{id}/entitlements`
  - `PATCH /identities/{id}/groups`
  - `PATCH /identities/{id}/roles`

Implementation and unit tests for these 3 will be provided right after in a new PR that will build on top of this.

## Changes
- introduced two interfaces, one for `IdentitiesService` and one for the `IdentitiesAuthorization` (out of scope for this PR)
- `service` interface renamed to `handler` to reflect its responsibilities
- introduced `ErrorResponseMapper` interface to allow product implementor to provide their own error to http response mapping
- custom errors now implement the `Error` interface and allow for custom message to be specified
  - added `ValidationError`
  - added `ErrorResponseMapper` constructor functions
  - existing tests adjusted to reflect changes
- added `writeResponse` function to avoid repetition
- added implementation for `/identities` endpoints handlers

## Tests
- all implemented handlers have corresponding tests
- missing tests for the 3 Patch endpoints not yet implemented


rebac-admin-backend/v1/response.go
```diff
@@ -25,47 +25,45 @@ func writeErrorResponse(w http.ResponseWriter, r *http.Request, err error) {
...
-		return true
-	case *resources.InvalidParamFormatError:
-		return true
-	case *resources.TooManyValuesForParamError:
-		return true
+// writeResponse is a helper method to avoid verbose repetition of very common instructions
+// it writes to the ResponseWriter either:
+// - the provided marshalledResponse if marshalErr is nil
+// - or the mapped error response based on marshalErr if it is not nil
+func writeResponse(w http.ResponseWriter, status int, marshalledResp []byte, marshalErr error) {
```
alesstimec, line 57: suggestion:
```
func writeResponse(w http.ResponseWriter, status int, responseObject interface{}) {
        data, err := json.Marshal(responseObject)
        if err != nil {
		writeErrorResponse(w, err)
		return
	}

	w.WriteHeader(status)
	_, err := w.Write(data)
	if err != nil {
		writeErrorResponse(w, err)
	}
}
```
which is a bit more general in that it removes the json marshalling step from all the functions that use it

alesstimec (commented):
error handling is a bit messy.. it needs cleaning up..

rebac-admin-backend/v1/response.go
BarcoMasile: Thanks Ales, I was actually undecisive about this and waited for review time to gather thoughts
I'll go that way, next revision coming soon

BarcoMasile:
> error handling is a bit messy.. it needs cleaning up..

Yes, we have it in our mind to improve error handling as well as providing basic "default" error -> api response for common errors so product developer are not _forced_ to provide their own to use the library. It'll be our focus after endpoints work.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
+// Copyright 2024 Canonical Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+package v1
+
+import (
+	"encoding/json"
+	"net/http"
+
+	r "github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
```
alesstimec, line 10: let's not do that.. it makes reading code more confusing than it needs to be.. just use resources.GetIdentitiesResponse etc..

alesstimec: also.. should the wireformats live in their own package so any potential clients can import them independently of the handler implementation package?

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+)
+
+// GetIdentities Get list of identities.
```
alesstimec, line 13: nitpick: GetIdentities returns the list of known identities.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+
+	response := r.GetIdentitiesResponse{
+		Data:   identities.Data,
+		Status: 200,
+	}
+
+	writeResponse(w, 200, response)
+}
+
+// PostIdentities Add an identity.
```
alesstimec, line 31: // PostIdentities adds a new identity.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+// (POST /identities)
+func (h handler) PostIdentities(w http.ResponseWriter, req *http.Request) {
+	identity := new(r.Identity)
+	defer req.Body.Close()
+
+	err := json.NewDecoder(req.Body).Decode(identity)
+	if err != nil {
+		writeErrorResponse(w, err)
```
alesstimec, line 39: see.. this is wrong.. writeErrorResponse will always return an internal server error status.. but in this case it's more of a "bad request", since we cannot decode it

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+
+	writeResponse(w, 200, response)
+}
+
+// PostIdentities Add an identity.
+// (POST /identities)
+func (h handler) PostIdentities(w http.ResponseWriter, req *http.Request) {
+	identity := new(r.Identity)
+	defer req.Body.Close()
+
```
alesstimec, line 36:
```
var identity resources.Identity
if err := json.NewDecoder(req.Body).Decode(&identity); err != nil {
.
.
```

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		response := h.IdentitiesErrorMapper.MapError(err)
+		writeResponse(w, response.Status, response)
+		return
+	}
+
+	writeResponse(w, 201, identity)
+}
+
+// DeleteIdentitiesItem Remove an identity.
```
alesstimec, line 53: // DeleteIdentities deletes the specified identity.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		response := h.IdentitiesErrorMapper.MapError(err)
+		writeResponse(w, response.Status, response)
+		return
+	}
+
+	w.WriteHeader(200)
+}
+
+// GetIdentitiesItem Get a single identity.
```
alesstimec, line 66: // GetIdentity returns the identity identified by the provided ID.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		response := h.IdentitiesErrorMapper.MapError(err)
+		writeResponse(w, response.Status, response)
+		return
+	}
+
+	writeResponse(w, 200, identity)
+}
+
+// PutIdentitiesItem Update an identity.
```
alesstimec, line 79: // UpdateIdentity updates the identity identified by the provided ID.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+	if err != nil {
+		response := h.IdentitiesErrorMapper.MapError(err)
+		writeResponse(w, response.Status, response)
+		return
+	}
+
+	writeResponse(w, 200, identity)
+}
+
+// GetIdentitiesItemEntitlements List entitlements the identity has.
```
alesstimec, line 106: // GetEntitlements returns the list of entitlements for an identity identified by the provided ID.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+
+	response := r.GetIdentityEntitlementsResponse{
+		Data:   entitlements.Data,
+		Status: 200,
+	}
+
+	writeResponse(w, 200, response)
+}
+
+// PatchIdentitiesItemEntitlements Add or remove entitlement to/from an identity.
```
alesstimec, line 124: // RemoveEntitlement removes an entitlement for an identity.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+// (PATCH /identities/{id}/entitlements)
+func (h handler) PatchIdentitiesItemEntitlements(w http.ResponseWriter, req *http.Request, id string) {
+	w.WriteHeader(http.StatusNotImplemented)
+}
+
+// GetIdentitiesItemGroups List groups the identity is a member of.
+// (GET /identities/{id}/groups)
+func (h handler) GetIdentitiesItemGroups(w http.ResponseWriter, req *http.Request, id string, params r.GetIdentitiesItemGroupsParams) {
```
alesstimec, line 132: // GetGroups returns the list of groups the identity is a member of.

we can keep the method names simple because we know this handler deals with identities. I would rather not put the http method into the method name.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+		Data:   groups.Data,
+		Status: 200,
+	}
+
+	writeResponse(w, 200, response)
+}
+
+// PatchIdentitiesItemGroups Add or remove the identity to/from a group.
+// (PATCH /identities/{id}/groups)
+func (h handler) PatchIdentitiesItemGroups(w http.ResponseWriter, req *http.Request, id string) {
```
alesstimec, line 150: perhaps we could name this method something like `UpdateGroups`?

alesstimec (commented):
I'd still like to suggest some changes to simplify error handling.. and a bit of renaming of handler methods to make them a bit simpler..

rebac-admin-backend/v1/identities.go
BarcoMasile: Unfortunately the name is the one autogenerated by the tool based on the API spec, so renaming methods is not possible unless modifying the OpenAPI spec with "extension properties" that relate to the specific tool we used.
I'm updateing the godocs but method names will unfortunately stay the same

rebac-admin-backend/v1/identities.go
BarcoMasile: see answer below

rebac-admin-backend/v1/identities.go
BarcoMasile: my bad, I'll fix this by passing a validation error which will translate to a bad request, thanks for spotting it!

rebac-admin-backend/v1/error.go
```diff
@@ -2,16 +2,104 @@
...
 
+import (
+	"fmt"
+
+	"github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
+)
+
 // UnauthorizedError represents unauthorized access error.
-type UnauthorizedError struct{}
+type UnauthorizedError struct {
```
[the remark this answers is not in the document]
BarcoMasile: As much as I like your idea I wouldn't make this change part of this PR. I would like to get the endpoints work going, and address your (much appreciated) suggestion a little later (next few days) so I have Babak support also

rebac-admin-backend/v1/identities.go
BarcoMasile: Wireformats are now in the same package but different file, we could move them to a different package, I'll track this so we can definitely reason about it. Thanks!

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+
+	r "github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
+)
+
+// GetIdentities Get list of identities.
+// (GET /identities)
+func (h handler) GetIdentities(w http.ResponseWriter, req *http.Request, params r.GetIdentitiesParams) {
+	identities, err := h.Identities.ListIdentities(&params)
+	if err != nil {
+		response := h.IdentitiesErrorMapper.MapError(err)
```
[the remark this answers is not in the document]
BarcoMasile: I'd fit this change into the above comment, so I'd leave the error handling improvements for another PR so I can get Babak's help.

rebac-admin-backend/v1/error.go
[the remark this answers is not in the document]
BarcoMasile: I tracked this in my notes.

alesstimec (approved):
ok, i've made myself heard.. i'll approve it, but i would still like @babakks to take a look at it.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+		writeResponse(w, response.Status, response)
+		return
+	}
+
+	response := resources.GetIdentitiesResponse{
+		Data:   identities.Data,
+		Status: 200,
```
babakks, line 25: Let's use `http.Status*` constants instead of integer literals.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+}
+
+// PostIdentities adds a new identity.
+// (POST /identities)
+func (h handler) PostIdentities(w http.ResponseWriter, req *http.Request) {
+	identity := new(resources.Identity)
+	defer req.Body.Close()
+
+	err := json.NewDecoder(req.Body).Decode(identity)
+	if err != nil {
```
babakks, line 38: As suggested by @alesstimec, it's best to limit the scope of the `err` temp variable to the `if`-statement.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+
+	"github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
+)
+
+// GetIdentities returns the list of known identities.
+// (GET /identities)
+func (h handler) GetIdentities(w http.ResponseWriter, req *http.Request, params resources.GetIdentitiesParams) {
+	identities, err := h.Identities.ListIdentities(&params)
+	if err != nil {
+		response := h.IdentitiesErrorMapper.MapError(err)
```
babakks, line 18: I'd apply @alesstimec suggestion in this PR. It's not a big change and I think it makes sense to be done here.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+	}
+
+	identity, err = h.Identities.CreateIdentity(identity)
+	if err != nil {
+		response := h.IdentitiesErrorMapper.MapError(err)
+		writeResponse(w, response.Status, response)
+		return
+	}
+
+	writeResponse(w, 201, identity)
```
babakks, line 50: The same thing here (i.e., `http.Status*` usage). I won't comment on the other instances.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+	writeResponse(w, 200, identity)
+}
+
+// PutIdentitiesItem updates the identity identified by the provided ID.
+// (PUT /identities/{id})
+func (h handler) PutIdentitiesItem(w http.ResponseWriter, req *http.Request, id string) {
+	identity := new(resources.Identity)
+	defer req.Body.Close()
+
+	err := json.NewDecoder(req.Body).Decode(identity)
```
babakks, line 85: Limit the `err` scope to the `if`-statement. Won't comment on the next occurrences.

rebac-admin-backend/v1/identities.go
```diff
@@ -0,0 +1,176 @@
...
+		Status: 200,
+	}
+
+	writeResponse(w, 200, response)
+}
+
+// PatchIdentitiesItemEntitlements Adds or removes entitlements to/from an identity.
+// (PATCH /identities/{id}/entitlements)
+func (h handler) PatchIdentitiesItemEntitlements(w http.ResponseWriter, req *http.Request, id string) {
+	w.WriteHeader(http.StatusNotImplemented)
```
babakks, line 127: Please create a card for the upcoming PR, and add a `// TODO([card-id])` comment here.

rebac-admin-backend/v1/identities.go
babakks: @alesstimec Yeah, the methods on this `handler` struct are internal. The product services only deal with the abstractions defined in the `interfaces` package.

rebac-admin-backend/v1/identities_test.go
```diff
@@ -0,0 +1,585 @@
...
+)
+
+var (
+	mockFirstName  string = "MockFirstName"
+	mockIdentityId string = "test-id"
+)
+
+//go:generate mockgen -package interfaces -destination ./interfaces/mock_identities.go -source=./interfaces/identities.go
+//go:generate mockgen -package v1 -destination ./mock_error_response.go -source=./error.go
+func TestHandler_GetIdentities(t *testing.T) {
```
babakks, line 30: Add a blank line before this, please.

rebac-admin-backend/v1/identities_test.go
```diff
@@ -0,0 +1,585 @@
...
+	mockWriter := httptest.NewRecorder()
+
+	sut := handler{Identities: mockIdentityService}
+	sut.GetIdentities(mockWriter, nil, mockParams)
+
+	result := mockWriter.Result()
+	defer result.Body.Close()
+
+	data, err := io.ReadAll(result.Body)
+	if err != nil {
```
babakks, line 56: Replace this and other similar assertions with:

```go
c.Assert(err, qt.IsNil)
```

There's no need to put the "Expected error to be nil, got %v" comment, because it's obvious. But, in case you needed to add an informative comment, you can use this:

```go
c.Assert(err, qt.IsNil, qt.Commentf("blah blah: %v", err))
```

rebac-admin-backend/v1/identities_test.go
```diff
@@ -0,0 +1,585 @@
...
+
+	c.Assert(result.StatusCode, qt.Equals, http.StatusOK)
+	c.Assert(response, qt.DeepEquals, &expectedResponse)
+}
+
+func TestHandler_PatchIdentitiesItemRolesSuccess(t *testing.T) {
+
+}
+
+func TestHandler_Failures(t *testing.T) {
```
babakks, line 369: I like this kind of table test. We should do the same thing for the happy path tests, and then merge them all in a single test function, because there's no point in having separate test functions.

rebac-admin-backend/v1/identities_test.go
```diff
@@ -0,0 +1,585 @@
...
+	}{
+		{
+			name: "TestGetIdentitiesFailure",
+			setupServiceMock: func(mockService *interfaces.MockIdentitiesService) {
+				mockService.EXPECT().ListIdentities(gomock.Any()).Return(nil, mockError)
+			},
+			triggerFunc: func(h handler, w *httptest.ResponseRecorder) {
+				h.GetIdentities(w, nil, mockParams)
+			},
+			skip: false,
```
babakks, line 396: You can just drop these `skip: false` assignments, due to defaults.

rebac-admin-backend/v1/identities_test.go
```diff
@@ -0,0 +1,585 @@
...
+
+	mockParams := resources.GetIdentitiesParams{}
+	mockErrorResponse := resources.Response{
+		Message: "mock-error",
+		Status:  http.StatusInternalServerError,
+	}
+
+	mockError := errors.New("test-error")
+
+	for _, test := range []struct {
```
babakks, line 382: Let's separate this table definition from the `for`-statement. It feels too inline being here.

rebac-admin-backend/v1/identities.go
babakks: @alesstimec The handlers are already implemented in the library's root package (i.e., not in the `resources` package).

babakks (approved):

babakks:
> ok, i've made myself heard.. i'll approve it, but i would still like @babakks to take a look at it.

Thanks for the review, @alesstimec. I put my comments.

rebac-admin-backend/v1/identities.go
BarcoMasile: Yes, this is taken care of in the next PR with the missing endpoints from this one
Thanks!

rebac-admin-backend/v1/identities_test.go
BarcoMasile: Will be in next review, thanks

rebac-admin-backend/v1/identities_test.go
BarcoMasile: These `skip` fields are dropped in the next PR with all endpoints implemented and tested

rebac-admin-backend/v1/identities_test.go
BarcoMasile: Done in the next revision coming up, thanks!

rebac-admin-backend/v1/interfaces/identities.go
```diff
@@ -0,0 +1,45 @@
...
+
+package interfaces
+
+import r "github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
+
+// IdentitiesService defines an abstract backend to handle Identities related operations.
+type IdentitiesService interface {
+
+	// ListIdentities returns a page of Identity objects of at least `size` elements if available
+	ListIdentities(params *r.GetIdentitiesParams) (*r.Identities, error)
```
babakks, line 12: One important thing that we missed, is to have a `context.Context` parameter for the methods defined in the interface.

babakks (changes requested):

rebac-admin-backend/v1/interfaces/identities.go
BarcoMasile: Yes, I took a note about this to discuss with you, but I'm happy to put it here in next review, coming up

babakks (approved):

PR #201, opened 2024-03-05
fix: refactor error types

babakks:
This PR applies some refactoring on error types. A new internal type, named `errorWithCode` is introduced, with a few exported helper methods to let users create errors.

Fixes CSS-7530

### DIFF
--- a/rebac-admin-backend/v1/error.go
+++ b/rebac-admin-backend/v1/error.go
@@ -53,6 +53,8 @@ func NewValidationError(message string) error {
 
 // ErrorResponseMapper is the basic interface to allow for error -> http response mapping
 type ErrorResponseMapper interface {
+	// MapError maps an error into a Response. If the method is unable to map the
+	// error (e.g., the error is unknown), it can just return nil.
 	MapError(error) *resources.Response
 }
 
--- a/rebac-admin-backend/v1/error.go
+++ b/rebac-admin-backend/v1/error.go
@@ -90,14 +90,12 @@ func NewDelegateErrorResponseMapper(m ErrorResponseMapper) ErrorResponseMapper {
 	return &delegateErrorResponseMapper{delegate: m}
 }
 
-// mapHandlerInternalError checks if the given error is an internal
-// handler-level error (i.e., an auto-generated error type) and return the
+// mapHandlerBadRequestError checks if the given error is an "Bad Request" error
+// thrown at the handler root (i.e., an auto-generated error type) and return the
 // equivalent errorWithStatus instance. If the given error is not an internal
 // handler error, this function will return nil.
-func mapHandlerInternalError(err error) *errorWithStatus {
-	// Note that these are all auto-generated error types, that should be
-	// regarded as bad request errors.
-	if !isHandlerInternalError(err) {
+func mapHandlerBadRequestError(err error) *errorWithStatus {
+	if !isHandlerBadRequestError(err) {
 		return nil
 	}
 	return &errorWithStatus{
@@ -106,7 +104,7 @@ func mapHandlerInternalError(err error) *errorWithStatus {
 	}
 }
 
-func isHandlerInternalError(err error) bool {
+func isHandlerBadRequestError(err error) bool {
 	switch err.(type) {
 	case *resources.UnmarshalingParamError:
 		return true
--- a/rebac-admin-backend/v1/error.go
+++ b/rebac-admin-backend/v1/error.go
@@ -8,46 +8,47 @@ import (
 	"github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
 )
 
-// UnauthorizedError represents unauthorized access error.
-type UnauthorizedError struct {
+// ErrorCode is a code which describes the class of error.
+type ErrorCode string
+
+const (
+	CodeUnauthorized    ErrorCode = "unauthorized"
+	CodeNotFound        ErrorCode = "not found"
+	CodeValidationError ErrorCode = "validation error"
+)
+
+type errorWithCode struct {
+	code    ErrorCode
 	message string
 }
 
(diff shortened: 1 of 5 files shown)


rebac-admin-backend/v1/error.go
```diff
@@ -8,46 +8,47 @@ import (
...
-}
+// ErrorCode is a code which describes the class of error.
+type ErrorCode string
 
-// NewUnauthorizedError returns a pointer to a new instance of UnauthorizedError with the provided message
-func NewUnauthorizedError(message string) *UnauthorizedError {
-	return &UnauthorizedError{message}
-}
+const (
+	CodeUnauthorized    ErrorCode = "unauthorized"
```
alesstimec, line 15: godocs, please

rebac-admin-backend/v1/error.go
```diff
@@ -8,50 +8,53 @@ import (
...
+	CodeNotFound        ErrorCode = "not found"
+	CodeValidationError ErrorCode = "validation error"
```
alesstimec, line 17: one thought.. if you weren't dead set on calling it a "validation error" we could just use http.Status* as error codes (which are just ints, if i recall corectly).. then you could also use http.StatusText to map error codes into text.. 
so instead of CodeUnauthorized, you'd use http.StatusUnauthorized, CodeNotFound -> http.StatusNotFound.. and CodeValidationError -> http.StatusBadRequest

alesstimec (approved):
LGTM.. with a suggestion to use http status codes

rebac-admin-backend/v1/response.go
```diff
@@ -26,6 +26,15 @@ func writeErrorResponse(w http.ResponseWriter, err error) {
...
 func mapErrorResponse(err error) *resources.Response {
+	// Theoretically, this should never happen, but we anyway have to check for
+	// a nil argument.
+	if err == nil {
+		return &resources.Response{
+			Status:  http.StatusOK,
+			Message: "",
+		}
+	}
+
```
BarcoMasile, line 37: nitpick(non-blocking): even though it doesn't hurt to check, being `mapErrorResponse` an unexported function we should not worry about err being nil

rebac-admin-backend/v1/error.go
```diff
@@ -53,6 +53,8 @@ func NewValidationError(message string) error {
 
 // ErrorResponseMapper is the basic interface to allow for error -> http response mapping
 type ErrorResponseMapper interface {
+	// MapError maps an error into a Response. If the method is unable to map the
+	// error (e.g., the error is unknown), it can just return nil.
```
BarcoMasile, line 57: suggestion: "it **must** return nil"
I think this is need because we rely on the return value to be nil in case "we" have to take care of the mapping

rebac-admin-backend/v1/error.go
BarcoMasile: suggestion: riding Ales' wave I'd like to suggest a change, since the `code` wording seems misleading:
- rename `code` to `status` (so we can also leverage `http.StatusText(code int)` to get the corresponding status string
- we add a `code int` or `statusCode int` to represent the actual http status code

This way we rely on http package for status strings and we only need to map a correct http code like 401 or 400 to build our errors.

The `Error()` function can then leverage the same fields as it does now, just with the renamed `code` into `status` field

BarcoMasile (approved):
LGTM, just a couple of comments with one suggestion on the ongoing discussion.

rebac-admin-backend/v1/response.go
babakks: Yeah, that's why I put the comment above it. A `nil` argument should never happen, but although this is an internal method, nevertheless, we need to do either of the following to fool-proof it: 
- Have tests to assert that `mapErrorResponse` is *never* called with a `nil` argument, which is hard to actually test.
- Accept that it's possible (though, very unlikely) and handle it like I did and have a test to verify this is the intended behavior.

rebac-admin-backend/v1/error.go
babakks: Thanks for the comment. I applied the changes in the latest commit.

rebac-admin-backend/v1/error.go
babakks: It's not there anymore. So I resolve this comment.

rebac-admin-backend/v1/error.go
```diff
@@ -4,49 +4,52 @@ package v1
...
-	CodeNotFound        ErrorCode = "not found"
-	CodeValidationError ErrorCode = "validation error"
-)
-
-type errorWithCode struct {
-	code    ErrorCode
+// errorWithStatus is an internal error representation that holds the corresponding
+// HTTP status code along with the error message.
+type errorWithStatus struct {
+	// status is the HTTP standard equivalent status status. Acceptable
```
BarcoMasile, line 15: typo: `status` is repeated

rebac-admin-backend/v1/error.go
```diff
@@ -87,11 +90,23 @@ func NewDelegateErrorResponseMapper(m ErrorResponseMapper) ErrorResponseMapper {
...
+// equivalent errorWithStatus instance. If the given error is not an internal
+// handler error, this function will return nil.
+func mapHandlerInternalError(err error) *errorWithStatus {
 	// Note that these are all auto-generated error types, that should be
 	// regarded as bad request errors.
+	if !isHandlerInternalError(err) {
+		return nil
+	}
+	return &errorWithStatus{
+		status:  http.StatusBadRequest,
```
BarcoMasile, line 104: question: internal error usually maps to 5xx, but this is for 400 bad requests, didn't the previous name of this function sound more appropriate?

alesstimec (approved):
Nice! thank you for this!

PR #204, opened 2024-03-06
feat: add `/groups` endpoints

babakks:
This PR implements handlers for `/groups` endpoints. It also fixes some minor issues with other endpoints' tests.

Fixes CSS-7255

### DIFF
--- a/rebac-admin-backend/v1/core.go
+++ b/rebac-admin-backend/v1/core.go
@@ -9,14 +9,24 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/interfaces"
 	"github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
 )
 
 // ReBACAdminBackendParams contains references to user-defined implementation
 // of required abstractions, called "backend"s.
 type ReBACAdminBackendParams struct {
-	GroupsService       GroupsServiceBackend
-	GroupsAuthorization GroupsAuthorizationBackend
+	IdentitiesService       interfaces.IdentitiesService
+	IdentitiesAuthorization interfaces.IdentitiesAuthorization
+	IdentitiesErrorMapper   ErrorResponseMapper
+
+	RolesService       interfaces.RolesService
+	RolesAuthorization interfaces.RolesAuthorization
+	RolesErrorMapper   ErrorResponseMapper
+
+	GroupsService       interfaces.GroupsService
+	GroupsAuthorization interfaces.GroupsAuthorization
+	GroupsErrorMapper   ErrorResponseMapper
 }
 
 // ReBACAdminBackend represents the ReBAC admin backend as a whole package.
--- a/rebac-admin-backend/v1/groups.go
+++ b/rebac-admin-backend/v1/groups.go
@@ -1,11 +1,241 @@
 // Copyright 2024 Canonical Ltd.
+// SPDX-License-Identifier: Apache-2.0
 
 package v1
 
-// GroupsServiceBackend defines an abstract backend to handle Groups.
-type GroupsServiceBackend interface {
+import (
+	"encoding/json"
+	"net/http"
+
+	"github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
+)
+
+// GetGroups returns the list of known groups.
+// (GET /groups)
+func (h handler) GetGroups(w http.ResponseWriter, req *http.Request, params resources.GetGroupsParams) {
+	ctx := req.Context()
+
+	groups, err := h.Groups.ListGroups(ctx, &params)
+	if err != nil {
+		response := h.GroupsErrorMapper.MapError(err)
+		writeResponse(w, response.Status, response)
+		return
+	}
+
+	response := resources.GetGroupsResponse{
+		Data:   groups.Data,
+		Status: http.StatusOK,
+	}
+
+	writeResponse(w, http.StatusOK, response)
 }
 
-// GroupsAuthorizationBackend defines an abstract backend to handle authorization for Groups.
-type GroupsAuthorizationBackend interface {
+// PostGroups adds a new group.
+// (POST /groups)
+func (h handler) PostGroups(w http.ResponseWriter, req *http.Request) {
+	ctx := req.Context()
+
+	group := new(resources.Group)
+	defer req.Body.Close()
+
+	if err := json.NewDecoder(req.Body).Decode(group); err != nil {
+		writeErrorResponse(w, NewValidationError("request doesn't match the expected schema"))
+		return
+	}
+
+	group, err := h.Groups.CreateGroup(ctx, group)
+	if err != nil {
+		response := h.GroupsErrorMapper.MapError(err)
+		writeResponse(w, response.Status, response)
+		return
+	}
+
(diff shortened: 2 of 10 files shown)


rebac-admin-backend/v1/core.go
```diff
@@ -9,14 +9,24 @@ import (
...
+	IdentitiesAuthorization interfaces.IdentitiesAuthorization
+	IdentitiesErrorMapper   ErrorResponseMapper
+
+	RolesService       interfaces.RolesService
+	RolesAuthorization interfaces.RolesAuthorization
+	RolesErrorMapper   ErrorResponseMapper
+
+	GroupsService       interfaces.GroupsService
+	GroupsAuthorization interfaces.GroupsAuthorization
+	GroupsErrorMapper   ErrorResponseMapper
```
BarcoMasile, line 29: suggestion: can we shorten the field names like in the `handler` struct?
example: `IdentitiesService` -> `Identities`

rebac-admin-backend/v1/identities.go
```diff
@@ -174,7 +174,7 @@ func (h handler) GetIdentitiesItemGroups(w http.ResponseWriter, req *http.Reques
 	writeResponse(w, http.StatusOK, response)
 }
 
-// PatchIdentitiesItemGroups adds or removes the identity to/from a group.
+// PatchIdentitiesItemGroups adds or removes the identity to/from groups.
```
BarcoMasile, line 177: praise: thanks for spotting this

rebac-admin-backend/v1/identities_test.go
BarcoMasile, line 1: praise: thanks for spotting these also

rebac-admin-backend/v1/interfaces/groups.go
```diff
@@ -0,0 +1,49 @@
...
+// GroupsService defines an abstract backend to handle Groups related operations.
+type GroupsService interface {
+
+	// ListGroups returns a page of Group objects of at least `size` elements if available.
+	ListGroups(ctx context.Context, params *resources.GetGroupsParams) (*resources.Groups, error)
+	// CreateGroup creates a single Group.
+	CreateGroup(ctx context.Context, group *resources.Group) (*resources.Group, error)
+
+	// GetGroup returns a single Group.
+	GetGroup(ctx context.Context, groupId string) (*resources.Group, error)
```
BarcoMasile, line 21: suggestion(proxied by Ales): a previous suggestion by Ales would have us append the string "identified by `groupId`", which I think it's better and I applied it to my godocs

BarcoMasile (approved):
LGTM, just a couple of nitpicks but also a few praises!

shipperizer (commented):
@babakks please do a rebase to drop [71af9af](https://github.com/canonical/identity-platform-admin-ui/pull/204/commits/71af9af603935b24d6717972c1410d2e08570980)

rebac-admin-backend/v1/groups.go
```diff
@@ -1,11 +1,241 @@
...
 }
 
-// GroupsAuthorizationBackend defines an abstract backend to handle authorization for Groups.
-type GroupsAuthorizationBackend interface {
+// GetGroupsItem returns the group identified by the provided ID.
+// (GET /groups/{id})
+func (h handler) GetGroupsItem(w http.ResponseWriter, req *http.Request, id string) {
+	ctx := req.Context()
+
+	identity, err := h.Groups.GetGroup(ctx, id)
```
alesstimec, line 76: s/identity/group/

rebac-admin-backend/v1/groups.go
```diff
@@ -1,11 +1,241 @@
...
+	}
+
+	if id != *group.Id {
+		writeErrorResponse(w, NewValidationError("group ID from path does not match the Group object"))
+		return
+	}
+
+	group, err := h.Groups.UpdateGroup(ctx, group)
+	if err != nil {
+		response := h.GroupsErrorMapper.MapError(err)
```
alesstimec, line 106: looking at the usage pattern, it seems better to have something like
```
h.GroupsErrorMapper.WriteErrorResponse(err)
return
```

rebac-admin-backend/v1/groups.go
```diff
@@ -1,11 +1,241 @@
...
+	groups, err := h.Groups.GetGroupIdentities(ctx, id, &params)
+	if err != nil {
+		response := h.GroupsErrorMapper.MapError(err)
+		writeResponse(w, response.Status, response)
+		return
+	}
+
+	response := resources.GetGroupIdentitiesResponse{
+		Data:   groups.Data,
+		Status: http.StatusOK,
```
alesstimec, line 171: this seems a bit unusual

rebac-admin-backend/v1/groups.go
```diff
@@ -1,11 +1,241 @@
...
+	roles, err := h.Groups.GetGroupRoles(ctx, id, &params)
+	if err != nil {
+		response := h.GroupsErrorMapper.MapError(err)
+		writeResponse(w, response.Status, response)
+		return
+	}
+
+	response := resources.GetIdentityRolesResponse{
+		Data:   roles.Data,
+		Status: http.StatusOK,
```
alesstimec, line 214: same here

alesstimec (approved):
lgtm with a few comments

rebac-admin-backend/v1/groups.go
babakks: Do you mean the status code being in the response type?

BarcoMasile: I think it's about the `.Data`.
Right now we are keeping this "wrapper" because we will need the backend service to pass (along with the data results) the paging info, since the controller layer doesn't have a way to know, for example, which will be the next page, the total items available, or the next page token in case of token-based pagination.
Right now we only use the `Data` field, but with pagination we'll have to rely on backend service layers to provide those other pieces of info.

rebac-admin-backend/v1/groups.go
BarcoMasile: I'm a little against this, reason being: I would let the mapper do the mapping work and that work only, following the single responsibility principle. What do you think?

babakks: Question; are you suggesting this to make it more readable/terse? or it's more about the semantics?

I'm asking this, because the `GroupsErrorMapper` is something that users (i.e., product services) should provide us with. So, changing `MapError` (which is expected to return a `Response` struct instance) to `WriteErrorResponse` (which implies user access to the undelrying `http.ResponseWriter`) affects the way we tend to abstract things. I, personally, rather not share the `http.ResponseWriter` instance with the user.

Anyway, if your suggestion is more about making it terse, I totally agree. I think we can add an internal helper method for this.

rebac-admin-backend/v1/core.go
babakks: I'm okay with that. Do you want to make the change here, or in another polish-like PR?

BarcoMasile: I would do it here but feel free to do it later
I'll link this comment in #205 also just to keep track

babakks: Done.

rebac-admin-backend/v1/groups.go
babakks: I think a generic helper method to do these two lines would be a good thing. I'll add it to my other PR (#205).

PR #208, opened 2024-03-07
feat: add `/entitlements` API set handler implementation

BarcoMasile:
## Description
This PR builds on top of the work done in #200 , and will need to be rebased on top of `feature-rebac-admin-ui-backend` branch after #207 gets merged: this PR real commits are only the last 3.

Implementations on the /capabilities endpoints handlers.

## Changes
- added implementation for mentioned endpoints
- added unit tests

### DIFF
--- a/rebac-admin-backend/v1/interfaces/entitlements.go
+++ b/rebac-admin-backend/v1/interfaces/entitlements.go
@@ -1,0 +1,21 @@
+// Copyright 2024 Canonical Ltd.
+// SPDX-License-Identifier: Apache-2.0
+
+package interfaces
+
+import (
+	"context"
+
+	"github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
+)
+
+type EntitlementsService interface {
+	// ListEntitlements returns the list of entitlements in JSON format.
+	ListEntitlements(ctx context.Context, params *resources.GetEntitlementsParams) ([]resources.EntityEntitlement, error)
+
+	// RawEntitlements returns the list of entitlements as raw text.
+	RawEntitlements(ctx context.Context) (*string, error)
+}
+
+type EntitlementsAuthorization interface {
+}
--- a/rebac-admin-backend/v1/interfaces/entitlements.go
+++ b/rebac-admin-backend/v1/interfaces/entitlements.go
@@ -14,7 +14,7 @@ type EntitlementsService interface {
 	ListEntitlements(ctx context.Context, params *resources.GetEntitlementsParams) ([]resources.EntityEntitlement, error)
 
 	// RawEntitlements returns the list of entitlements as raw text.
-	RawEntitlements(ctx context.Context) (*string, error)
+	RawEntitlements(ctx context.Context) (string, error)
 }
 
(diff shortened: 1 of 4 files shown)


rebac-admin-backend/v1/interfaces/capabilities.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"context"
+
+	"github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
+)
+
+// CapabilitiesService defines an abstract backend to handle capabilities related operations.
+type CapabilitiesService interface {
+	ListCapabilities(ctx context.Context) (*resources.Capabilities, error)
```
babakks, line 14: We don't need an interface for the `/capabilities` endpoint. This one endpoint should be handled by the library (based on non-`nil` service backends provided by the product services).

rebac-admin-backend/v1/interfaces/entitlements.go
```diff
@@ -0,0 +1,21 @@
...
+
+	"github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
+)
+
+type EntitlementsService interface {
+	// ListEntitlements returns the list of entitlements in JSON format.
+	ListEntitlements(ctx context.Context, params *resources.GetEntitlementsParams) ([]resources.EntityEntitlement, error)
+
+	// RawEntitlements returns the list of entitlements as raw text.
+	RawEntitlements(ctx context.Context) (*string, error)
```
babakks, line 17: Since this is a primitive return type, we can just use the value type.

babakks (commented):
LGTM, except for the `/capabilities` endpoint.

rebac-admin-backend/v1/interfaces/entitlements.go
BarcoMasile: I made it a pointer so it can be nillable in case of an error is present, but I think it's safe to make it concrete and be consistent

BarcoMasile: Done with the new commit, thanks!

rebac-admin-backend/v1/interfaces/capabilities.go
babakks: Yeah, I thought of having fine-grained capabilities support before, but that seemed a bit irrelevant. However, your example is not that far from reality. So, let's rephrase my comment. Let's provide some mechanism for the product service to inform us on the capabilities, and then we translate it to the OpenAPI spec response in the handler. How about that?

BarcoMasile: I think the simplest way to do that is using the same interface approach we used until now.
That is in all effect the mechanism you refer to, since it lets product teams provide that info. Any other solution would probably be "over-designing" at this point.

babakks (approved):

rebac-admin-backend/v1/interfaces/capabilities.go
babakks: I resolve this comment, as we decided to merge it as is and polish it later, with more context/ideas from other.

PR #209, opened 2024-03-07
fix: polish writing service error responses

babakks:
This PR polished the way we map errors thrown by services (e.g., `IdentitiesService`) to the OpenAPI spec response type.

This PR also:
- Removes the unused `delegateErrorResponseMapper` type and applies the same approach within the `mapServiceErrorResponse` function.
- Drops unused `skip` variable from table tests.
- Fixes `handler` struct instantiation by providing it with all services. 
- Adds explicit `Content-Type: application/json` header to responses.
- Improve some field names by removing the `Service` suffix.

Fixes CSS-7635

### DIFF
--- a/rebac-admin-backend/v1/response.go
+++ b/rebac-admin-backend/v1/response.go
@@ -63,3 +63,27 @@ func writeResponse(w http.ResponseWriter, status int, responseObject interface{}
 		writeErrorResponse(w, err)
 	}
 }
+
+// mapServiceErrorResponse maps errors thrown by services to the designated
+// response type. If the given mapper is nil, the method uses the default
+// mapping strategy.
+//
+// This method should never returns nil response.
+func mapServiceErrorResponse(mapper ErrorResponseMapper, err error) *resources.Response {
+	var response *resources.Response
+	if mapper != nil {
+		response = mapper.MapError(err)
+	}
+
+	if response == nil {
+		response = mapErrorResponse(err)
+	}
+	return response
+}
+
+// writeServiceErrorResponse is a helper method that maps errors thrown by
+// services and writes them to the HTTP response stream.
+func writeServiceErrorResponse(w http.ResponseWriter, mapper ErrorResponseMapper, err error) {
+	response := mapServiceErrorResponse(mapper, err)
+	writeResponse(w, response.Status, response)
+}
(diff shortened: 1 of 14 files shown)


alesstimec (approved):
Yes! Thank you! Much cleaner!

rebac-admin-backend/v1/response.go
```diff
@@ -56,10 +58,35 @@ func writeResponse(w http.ResponseWriter, status int, responseObject interface{}
...
+
+	if response == nil {
+		response = mapErrorResponse(err)
 	}
+	return response
+}
+
+// writeServiceErrorResponse is a helper method that maps errors thrown by
+// services and writes them to the HTTP response stream.
+func writeServiceErrorResponse(w http.ResponseWriter, mapper ErrorResponseMapper, err error) {
```
BarcoMasile, line 85: honestly, I'm not a fan of this since Mapping and Writing IMHO should not live under the same roof, but I'll step aside on this one

rebac-admin-backend/v1/response.go
```diff
@@ -56,10 +58,35 @@ func writeResponse(w http.ResponseWriter, status int, responseObject interface{}
...
+	response := mapServiceErrorResponse(mapper, err)
+	writeResponse(w, response.Status, response)
+}
+
+func setJSONContentTypeHeader(w http.ResponseWriter) {
```
BarcoMasile, line 90: this is something that would usually be done in a middleware. Specifically, `chi-server` in my experience seems to disregard this unless it's done in a middleware, but happy to be proven wrong
I think we need a POC for doubts like this

BarcoMasile (approved):
approved with a couple of considerations

rebac-admin-backend/v1/response.go
babakks: I know what you say, but note that this is a helper method to avoid repeating ourselves, and of course, ensure we're not missing the steps.

rebac-admin-backend/v1/response.go
babakks: Basically, the `w.Write` takes care of setting the content-type header (as of docs, it reads the first 512 bytes of the body and decides if it's a JSON or something else). I just don't want to rely on it and it's okay to set it by the response writer functions (because there we explicitly do the JSON marshalling and are sure that the response is a JSON). And also note that, the response is not always a JSON:

https://github.com/canonical/identity-platform-admin-ui/blob/52dc995436f2e0be775b3bf47ad2cecb690731ac/rebac-admin-backend/v1/response.go#L17-L22

As for the middleware, I think keeping this as simple as this is okay. If we really needed a middleware approach we can always refactor.

PR #210, opened 2024-03-07
feat: add `/swagger.json` handler

babakks:
This PR adds the `/swagger.json` handler.

Fixes CSS-7326

### DIFF
--- a/rebac-admin-backend/v1/swagger.go
+++ b/rebac-admin-backend/v1/swagger.go
@@ -14,7 +14,7 @@ import (
 func (h handler) SwaggerJson(w http.ResponseWriter, req *http.Request) {
 	swagger, err := resources.GetSwagger()
 	if err != nil {
-		writeErrorResponse(w, NewUnknownError("cannot retrieve swagger instance"))
+		writeErrorResponse(w, NewUnknownError("cannot retrieve swagger data"))
 		return
 	}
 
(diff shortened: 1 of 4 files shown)


rebac-admin-backend/v1/swagger.go
```diff
@@ -0,0 +1,28 @@
...
+
+	"github.com/canonical/identity-platform-admin-ui/rebac-admin-backend/v1/resources"
+)
+
+// Returns the OpenAPI spec as a JSON file.
+// (GET /swagger.json)
+func (h handler) SwaggerJson(w http.ResponseWriter, req *http.Request) {
+	swagger, err := resources.GetSwagger()
+	if err != nil {
+		writeErrorResponse(w, NewUnknownError("cannot retrieve swagger instance"))
```
alesstimec, line 17: suggestion: s/instance/data/

rebac-admin-backend/v1/swagger.go
```diff
@@ -0,0 +1,28 @@
...
+		return
+	}
+
+	body, err := swagger.MarshalJSON()
+	if err != nil {
+		writeErrorResponse(w, NewUnknownError("cannot marshal spec as JSON"))
+		return
+	}
+
+	w.Write(body)
```
alesstimec, line 27: would we want to log in case Write returns an error.. :shrug:

alesstimec (approved):
LGTM with two suggestions

rebac-admin-backend/v1/swagger.go
babakks: Actually I had that in my mind, but we haven't yet selected a logging package/implementation for the library. I think for now, I can just use the standard logger and we can change it, if needed, in the future.

BarcoMasile (approved):
LGTM

PR #212, opened 2024-03-13
Update setup docs

huwshimi:
## Description

~~I've added the address to the port forward config so that the API port can be accessed from outside the host (so that the API can be run in a Multipass and accessed from outside).~~ I've also added a couple of steps that weren't listed in the setup instructions.

## Changes
- Add additional setup instructions.
- ~~Expose the API port.~~

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,9 +44,15 @@ As a requirement, please make sure to:
       microk8s enable registry
       # ensure kubectl is configured to use microk8s
       microk8s.kubectl config view --raw > $HOME/.kube/config
+      # Alias kubectl so that it can be used by Skaffold
+      snap alias microk8s.kubectl kubectl
 
 
-* ensure [`skaffold`](https://github.com/GoogleContainerTools/skaffold), [`container-structure-test`](https://github.com/GoogleContainerTools/container-structure-test) and [`docker`](https://docs.docker.com/engine/install/ubuntu/) are installed according to their documentation
+* ensure [`skaffold`](https://github.com/GoogleContainerTools/skaffold), [`container-structure-test`](https://github.com/GoogleContainerTools/container-structure-test), [`docker`](https://docs.docker.com/engine/install/ubuntu/) and [helm](https://helm.sh/docs/intro/install) are installed according to their documentation
+
+* initialise LXD so that it has a default profile when used by `rockcraft` during `skaffold build`
+
+      sudo lxd init --auto
 
 Run `make dev` to get a working environment in k8s
 


BarcoMasile:
FYI: I think you're destination branch is wrong, you probably wanted to set main, since the `feature-rebac-admin-ui-backend` branch is for a library that doesn't affect skaffold.

skaffold.yaml
```diff
@@ -61,6 +61,7 @@ portForward:
   namespace: default
   port: 80
   localPort: 8000
+  address: 0.0.0.0
```
shipperizer, line 64: isn't this something that should be done on multipass/VM setup?
reason is this might clash with something else running on the machine if you are using the main host compared to a VM

README.md
```diff
@@ -44,9 +44,14 @@ As a requirement, please make sure to:
...
       microk8s.kubectl config view --raw > $HOME/.kube/config
+      snap alias microk8s.kubectl kubectl
 
 
-* ensure [`skaffold`](https://github.com/GoogleContainerTools/skaffold), [`container-structure-test`](https://github.com/GoogleContainerTools/container-structure-test) and [`docker`](https://docs.docker.com/engine/install/ubuntu/) are installed according to their documentation
+* ensure [`skaffold`](https://github.com/GoogleContainerTools/skaffold), [`container-structure-test`](https://github.com/GoogleContainerTools/container-structure-test), [`docker`](https://docs.docker.com/engine/install/ubuntu/) and [helm](https://helm.sh/docs/intro/install) are installed according to their documentation
+
+* initialise LXD
+
+      sudo lxd init --auto
```
shipperizer, line 54: not sure i understand why lxd is part of this, but if it's for troubleshooting when u have docker and lxd running together we'd better have a separate section

shipperizer (changes requested):
@huwshimi try to follow the conventional commit standard https://www.conventionalcommits.org/en/v1.0.0/

feel free to rewrite the commit message and use the `docs: ` or `ci: ` prefix

also see the comment from @BarcoMasile re the branch

skaffold.yaml
huwshimi: No worries, I'll revert this change. It would be nice if this could be set via an environment variable or flag, but I can't find an obvious way that skaffold supports this.

README.md
huwshimi: LXD is used by rockcraft and without a default profile you get an error when running `make dev`:

```
Command standard error output: b'Error: Failed instance creation
: Failed creating instance record: Failed initialising instance: Failed getting root disk:
 No root device could be found\n'
```

huwshimi (commented):
I've rebased onto main, removed the exposed port and fixed the commit message.

README.md
shipperizer: aaah my bad, that is needed when we build the image

fair point, apologies, do u mind adding something along these lines?

```suggestion
* initialise LXD to use `rockcraft` during `skaffold build`

      sudo lxd init --auto
```

shipperizer (approved):
see if u can add the tiny extra bit re rockcraft

anyway it's good to go

README.md
huwshimi: Done, I've also added a comment to explain why we need to alias kubectl.

shipperizer (approved):

PR #220, opened 2024-03-15
fix/noop ofga

shipperizer:
- **fix: pass interface to roles API to allow for openfga noop client**
- **fix: adjust openfga NoopClient setup**


BarcoMasile (approved):

shipperizer:
Creating  #221 and #222 as follow ups

PR #224, opened 2024-03-18
feat: add separate CI worflows to run library tests and linting

BarcoMasile:
## Description
This PR creates a separate CI workflow for the library folder `reback-admin-backend`.
It performes linting and runs unit tests. Right now the action runs on both merges on branch `feature-reback-admin-ui-backend` and on all opened pull requests. When the folder will be extracted only filenames and relative usages of their path will need to be changed.

## Changes
For the linting, I used [golangci-lint](https://github.com/marketplace/actions/run-golangci-lint) github action with `latest` version and the `working-directory` set to the library folder `rebac-admin-backend`.
For unit tests, I just run `make test`.


.github/workflows/ci-rebac-admin-backend.yaml
```diff
@@ -0,0 +1,19 @@
...
+    branches:
+      - "*"
+
+jobs:
+  lint:
+    name: Perform linting
+    uses: ./.github/workflows/rebac-admin-backend-lint.yaml
+  unit-test:
+    name: Run unit tests
+    uses: ./.github/workflows/rebac-admin-backend-unittest.yaml
```
canonical-iam, line 19: unless we reuse this kind of stuff multiple times, i d say simply put it inline and blend all the files into one

.github/workflows/rebac-admin-backend-unittest.yaml
```diff
@@ -0,0 +1,29 @@
...
+
+jobs:
+  unit-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 # v4
+
+      - uses: actions/setup-go@93397bea11091df50f3d7e59dc26a7711a8bcfbe # v4
+        with:
+          go-version: '1.21'
```
babakks, line 14: The same comment.

rebac-admin-backend/Makefile
```diff
@@ -22,5 +22,9 @@ mocks:
 
 # Run tests
 test: mocks
-	go test ./...
+	go test ./... -cover -coverprofile coverage_source.out
```
babakks, line 25: Since we're using this Makefile target in CI, we should be able to disable caching. The idiomatic way to do this is to call `go test` with `-count=1` option.

In the Makefile, we can do it like this:

```Makefile
test: mocks
    go test ./... -cover -coverprofile coverage_source.out $(ARGS)
    # ...
```

And run it in the CI with:

```
make ARGS="-count=1" test
```

rebac-admin-backend/Makefile
```diff
@@ -22,5 +22,9 @@ mocks:
...
+	# this will be cached, just needed to the test.json
```
babakks, line 26: Did you mean "to *get* the test.json"?

rebac-admin-backend/Makefile
```diff
@@ -22,5 +22,9 @@ mocks:
 
 # Run tests
 test: mocks
```
babakks, line 24: I'd rather have two Makefile targets, one for coverage and another for just running the tests.

rebac-admin-backend/Makefile
```diff
@@ -22,5 +22,9 @@ mocks:
...
-	go test ./...
+	go test ./... -cover -coverprofile coverage_source.out
+	# this will be cached, just needed to the test.json
+	go test ./... -cover -coverprofile coverage_source.out -json > test_source.json
+	cat coverage_source.out | grep -v "mock_*" | tee coverage.out
+	cat test_source.json | grep -v "mock_*" | tee test.json
```
babakks, line 29: Since we can also run this target on our local machines, we should add `test_source.json`, `coverage_source.out`, `coverage.out` and `test.json` to the `.gitignore` file.

babakks (approved):

rebac-admin-backend/Makefile
BarcoMasile: The *.out files are already in the .gitignore
I'll add the other ones, thanks

rebac-admin-backend/Makefile
BarcoMasile: I'll add a target that only does testing for local development purposes.
For the CI, I'd leave the coverage in if you agree

rebac-admin-backend/Makefile
BarcoMasile: Great! I'll add this right away

rebac-admin-backend/Makefile
babakks: That's the point. Thanks.

babakks (approved):

PR #226, opened 2024-03-18
Add identity providers to UI

edlerd:
# Done

- add idp management to ui
- add logo to ui
- simplify dev setup
- update readme
- update dependencies
- move creation/edit of providers and idps to side panels

# Screenshots

Screenshot from 2024-03-22 18-15-53
Screenshot from 2024-03-22 18-15-46
Screenshot from 2024-03-22 18-15-27
Screenshot from 2024-03-22 18-14-32
Screenshot from 2024-03-22 18-14-20
Screenshot from 2024-03-22 18-14-08
Screenshot from 2024-03-22 18-13-58
Screenshot from 2024-03-22 18-13-34
Screenshot from 2024-03-22 18-13-29
Screenshot from 2024-03-22 18-12-46


ui/src/types/provider.d.ts
```diff
@@ -0,0 +1,18 @@
+export interface IdentityProvider {
```
BarcoMasile, line 1: nitpick: these fields are all non-mandatory right now, according to the spec, so you would probably want to put them as optional

build.sh
```diff
@@ -10,7 +10,7 @@ set -e
 
 rockcraft pack -v
 
-skopeo --insecure-policy \
+/snap/rockcraft/current/bin/skopeo --insecure-policy \
```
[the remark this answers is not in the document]
edlerd: I encountered issues with the version of skopeo available with `apt install skopeo` being too old. As rockcraft ships with skopeo, I thought we could directly use that version. Hence also removing the note from readme to install skopeo, as it can be simplified by using the skopeo shipped with rockcraft.

ui/src/types/provider.d.ts
edlerd: Good point, thanks :)
Will change it.

ui/src/pages/providers/ProviderCreate.tsx
```diff
@@ -0,0 +1,90 @@
...
+  });
+
+  const formik = useFormik<ProviderFormTypes>({
+    initialValues: {
+      client_secret: "secret-9",
+      id: "okta_347646e49b484037b83690b020f9f629",
+      client_id: "347646e4-9b48-4037-b836-90b020f9f629",
+      provider: "okta",
+      mapper: "file:///etc/config/kratos/okta_schema.jsonnet",
+      scope: "email",
```
BarcoMasile, line 34: issue: hardcoding values, **especially** `client-secret` should not appear in any frontend code to avoid leaks

edlerd: Right, those values should all be empty string. Those have values atm to ease testing.

BarcoMasile: Totally agree, thank you!

build.sh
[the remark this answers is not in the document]
shipperizer: that is fair but at the same time you can solve the problem with a symlink

`ln -s /snap/rockcraft/current/bin/skopeo /usr/local/bin/skopeo`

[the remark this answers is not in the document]
edlerd: True, I'll add this to the readme then instead and revert this change.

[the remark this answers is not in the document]
edlerd: reverted this change and added it to the readme

ui/src/pages/providers/ProviderCreate.tsx
edlerd: Removed the defaults here.

BarcoMasile (approved):

shipperizer (approved):
@edlerd looks good, do u mind dropping some screenshots in the PR description?

goodnto merge anyway

edlerd:
> @edlerd looks good, do u mind dropping some screenshots in the PR description?
> 
> goodnto merge anyway

@shipperizer thanks for the review. I added a bunch of screens to the description.

PR #232, opened 2024-03-20
IAM 366: Groups API

shipperizer:
IAM-366: groups API following openapi spec v0.8.0 (to a degree)

- **feat: handlers for groups API**
- **feat: create groups service**
- **feat: introduce groups API converter to deal with authorization in the middleware**

### DIFF
--- a/internal/authorization/converters.go
+++ b/internal/authorization/converters.go
@@ -204,3 +204,95 @@ func (c RoleConverter) Map(r *http.Request) []Permission {
 		{Relation: relation(r), ResourceID: fmt.Sprintf("%s:%s", c.TypeName(), role_id)},
 	}
 }
+
+// GET /api/v0/groups
+// GET /api/v0/groups/{id}
+// POST /api/v0/groups
+// PATCH /api/v0/groups/{id}
+// DELETE /api/v0/groups/{id}
+// GET /api/v0/groups/{id}/roles
+// POST /api/v0/groups/{id}/roles
+// DELETE /api/v0/groups/{id}/roles/{r_id}
+// GET /api/v0/groups/{id}/entitlements
+// POST /api/v0/groups/{id}/entitlements
+// DELETE /api/v0/groups/{id}/entitlements/{e_id}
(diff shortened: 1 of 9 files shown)


nsklikas (approved):

internal/authorization/converters.go
```diff
@@ -204,3 +204,95 @@ func (c RoleConverter) Map(r *http.Request) []Permission {
...
+			{
+				Relation:   CAN_EDIT,
+				ResourceID: fmt.Sprintf("%s:%s", c.TypeName(), group_id),
+			},
+		}
+	}
+
+	// DELETE /api/v0/groups/{id}/entitlements
+	// POST /api/v0/groups/{id}/entitlements
+	if strings.HasSuffix(r.URL.Path, "entitlements") && (r.Method == http.MethodDelete || r.Method == http.MethodPost) {
```
BarcoMasile, line 264: nitpick: `/entitlements` suffixed endpoints for group don't support POST operations in this PR, only GET, PATCH and DELETE

internal/authorization/converters.go
```diff
@@ -204,3 +204,95 @@ func (c RoleConverter) Map(r *http.Request) []Permission {
...
+// GET /api/v0/groups/{id}/identities
+// PATCH /api/v0/groups/{id}/identities
+// DELETE /api/v0/groups/{id}/identities/{i_id}
+type GroupConverter struct{}
+
+func (c GroupConverter) TypeName() string {
+	return GROUP_TYPE
+}
+
+func (c GroupConverter) Map(r *http.Request) []Permission {
```
BarcoMasile, line 228: question(possible issue): one thing we don't do here is actually check for the first `id` (`group_id`) to be non-blank, may raise issues at runtime if we don't validate endpoints pathvariables before this middleware is run. Or maybe `chi` does it on its own?

Could it be beneficial to add as a first check after parameter retrieval if the `group_id` is empty and the method allows for an empty parameter, like GET ( -> `can_view`) or POST ( -> `can_edit`)?

PR #234, opened 2024-03-20
feat: add validation for `/groups` endpoints

babakks:
This PR adds validation for `/groups` endpoints to the ReBAC Admin UI library. This functionality is implemented as a decorated handler that wraps/decorates the actual handler and applies validation logic.

For now, this is only done for `/groups` endpoints. After this has landed, follow up PRs would be submitted for the rest of endpoints.

Note that the bulk of methods in `rebac-admin-backend/v1/validation.go` are there to implement the interface. I could've embedded the `Unimplemented` struct for this, but since these were already there, I didn't change it.

Fixes CSS-7621


rebac-admin-backend/v1/validation.go
```diff
@@ -0,0 +1,245 @@
...
+// request body and if it's okay, will delegate to the provided callback with a
+// new HTTP request instance with the parse body in the context.
+func setRequestBodyInContext[T any](w http.ResponseWriter, r *http.Request, f func(w http.ResponseWriter, r *http.Request, body *T)) {
+	body, err := parseRequestBody[T](r)
+	if err != nil {
+		writeErrorResponse(w, err)
+		return
+	}
+	f(w, newRequestWithBodyInContext(r, body), body)
+}
```
shipperizer, line 73: this looks really smart, but i m finding it really hard to understand

the use of generics could be dropped if u offload [this](https://github.com/canonical/identity-platform-admin-ui/pull/234/files#diff-17dd23c5e4825273ddfc3f074394963daa15bc9a48de3e41d1d9433e6932e7f7R54) to each implementation in [here]https://github.com/canonical/identity-platform-admin-ui/pull/234/files#diff-1088fa4484e7da34fc7f69eee616d0f5ad686a02941cd9f6c0b28d737c1a8cbeR19]

while for `validation_utils.go` generics make a lot of sense, here i m not 100% sure, especially if we read the sections in https://go.dev/blog/when-generics on `when not to use`

rebac-admin-backend/v1/groups_validation_types.go
```diff
@@ -0,0 +1,70 @@
...
+	}
+	return nil
+}
+
+func validateGroupIdentitiesPatchRequestBody(value *resources.GroupIdentitiesPatchRequestBody) error {
+	if len(value.Patches) == 0 {
+		return NewRequestBodyValidationError("empty patch array")
+	}
+	return validateSlice[resources.GroupIdentitiesPatchItem](value.Patches, validateGroupIdentitiesPatchItem)
+}
```
shipperizer, line 49: this could be implemented on types generation, see https://github.com/deepmap/oapi-codegen/issues/227

basically using the [extensions](https://github.com/deepmap/oapi-codegen?tab=readme-ov-file#extensions) in the specific the `x-oapi-codegen-extra-tags`, this would add an extra yaml file, but you would have validation tags/annotations in the struct itself

rebac-admin-backend/v1/validation.go
babakks: I'm not sure if I'm getting your point correctly, but I don't have a strong opinion on using generics for this part. Of course, parsing/validation step can be done without generics. For example, without generics the handler methods in the `handlerWithValidation` struct would look change from this:

```go
func (v handlerWithValidation) PostGroups(w http.ResponseWriter, r *http.Request) {
  setRequestBodyInContext[resources.Group](w, r, func(w http.ResponseWriter, r *http.Request, body *resources.Group) {
    if err := validateGroup(body); err != nil {
      writeErrorResponse(w, err)
      return
    }
    v.handler.PostGroups(w, r)
  })
}
```

to this:

```go
func (v handlerWithValidation) PostGroups(w http.ResponseWriter, r *http.Request) {
  body := &resources.Group{}
  setRequestBodyInContext(body, w, r, func(w http.ResponseWriter, r *http.Request) {
    if err := validateGroup(body); err != nil {
      writeErrorResponse(w, err)
      return
    }
    v.handler.PostGroups(w, r)
  })
}
```

Is this what you're asking for?

rebac-admin-backend/v1/groups_validation_types.go
babakks: Yeah, I checked the extensions, but what I don't like about them is that they clutter the spec with generator-specific tags.

rebac-admin-backend/v1/groups_validation.go
BarcoMasile, line 1: My issue with this approach is that it is not a real decorator (which would avoid repetition), but it is a full on proxy, since a new implementation of all handler methods is required to do it.
I think we could do validation in one single place, thanks to the fact that we already have specific structs for each request body types (see links for examples on my proposal).

#### Proposal
Leveraging [validator](https://github.com/go-playground/validator) library, and a single shared `var validate *validator.Validate` instance (since it is thread-safe, [link](https://github.com/go-playground/validator/issues/315#issuecomment-336145180)), we could use [`RegisterStructValidation`](https://github.com/go-playground/validator/blob/master/_examples/struct-level/main.go#L80) method to have our [custom validation functions](https://github.com/go-playground/validator/blob/master/_examples/struct-level/main.go#L162-L172) for all the typed request bodies (as you did), without the need to add field level tags, so no need to touch generated files), just by running [`validate.Struct(requestBody)`](https://github.com/go-playground/validator/blob/master/_examples/struct-level/main.go#L111).

After this setup, a single middleware would then be able to grab the request body (`if r.method != http.MethodGet` for example) and run `validator.Validate` on the specific type, it would just need to choose the right type based on the URL and method.
If using a single middleware makes type guards too long in the code, then we could have one middleware per resource type, but still the whole solution would be much more stramlined and simple.

rebac-admin-backend/v1/validation.go
shipperizer: the 2 bits of code above are basically the same, if we can achieve something without generics we should

it arguably makes the code harder to read, it doesn't speed it up and doesn't really bring anything to the plate apart from sugar syntax in this case

rebac-admin-backend/v1/groups_validation_types.go
shipperizer: that is fair, but at the same time right now we have types without effective embedded validation which doesn't sound ideal either

rebac-admin-backend/v1/validation.go
babakks: @alesstimec Yeah, I've seen generics used just for ensuring type safety and helping with type casting in other legit OSS projects (other languages).

rebac-admin-backend/v1/groups_validation_types.go
babakks: That's fair, too. But the solution would be to change the code generator instead of touching the spec.

rebac-admin-backend/v1/response.go
```diff
@@ -60,7 +66,10 @@ func writeResponse(w http.ResponseWriter, status int, responseObject interface{}
 
 	setJSONContentTypeHeader(w)
 	w.WriteHeader(status)
-	w.Write(data)
+	if _, err := w.Write(data); err != nil {
+		// TODO(CSS-7642): we should log the error.
```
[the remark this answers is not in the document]
babakks: In that card (CSS-7642) we are going to add logging, which should support an external logger behind the standard `log.Logger` interface. That's why I left the implementation for now.

rebac-admin-backend/v1/validation.go
```diff
@@ -0,0 +1,245 @@
...
+
+// GetIdentityProviders validates request body for the GetIdentityProviders method and delegates to the underlying handler.
+func (v handlerWithValidation) GetIdentityProviders(w http.ResponseWriter, r *http.Request, params resources.GetIdentityProvidersParams) {
+	v.handler.GetIdentityProviders(w, r, params)
+}
+
+// PostIdentityProviders validates request body for the PostIdentityProviders method and delegates to the underlying handler.
+func (v handlerWithValidation) PostIdentityProviders(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
babakks: You're right about the embedded struct. In the next push, I'll embed the handler so we only implement the methods we need.

rebac-admin-backend/v1/groups_validation.go
babakks: I actually went that way a bit (except for making it as a middleware, more on that later) and faced some issues. The most obvious one is `validator` module is not meant to be used like this. It depends on/works with tags, and lets you have your own custom tags to implement more complex validations. I tried to use it without tags and implemented a few struct-level validators but then I noticed there's no decent error reporting mechanism for un-tagged usages. For example, this is the official example:

```go
func UserStructLevelValidation(sl validator.StructLevel) {

	user := sl.Current().Interface().(User)

	if len(user.FirstName) == 0 && len(user.LastName) == 0 {
		sl.ReportError(user.FirstName, "fname", "FirstName", "fnameorlname", "")
		sl.ReportError(user.LastName, "lname", "LastName", "fnameorlname", "")
	}

	// plus can do more, even with different tag than "fnameorlname"
}
```

Here, the `sl.ReportError` method has the following signature:

```go
ReportError(field interface{}, fieldName, structFieldName string, tag, param string)
```

As you see, it expects a tag reference. I, anyway, made a custom validator for `Group` structs and passed the string `"cannot-be-empty"` as the tag (when `Group.Name` is empty). When I tested the error message was:

```
Key: 'Group.name' Error:Field validation for 'name' failed on the 'cannot-be-empty' tag
```

Which is not accurate (because there's no `cannot-be-empty` tag around) and also not that straightforward/clear for the end user. I'm not just complaining about the error message formatting here, but my point is the library is not meant to be used like this.

babakks: As for the middleware suggestion, I'm against it because that means we should manually cover all the endpoints and duplicate something like the auto-generated routing. I want to use the generated code as much as possible and avoid repeating it. That's why I had to go with a decorated handler here (which uses the routing done in the auto-generated code).

Regarding this not being a true decorator, I don't think that's the case. I mean, with using an embedded struct as you suggested, the unnecessary repetitions would fade out. But the thing is validation, in essence, should apply on every request. You already pointed out that for some endpoints we need to check the *absence* of a request body. This means, all endpoints are actually meant to be implemented. But, for now we can skip those.

rebac-admin-backend/v1/validation.go
```diff
@@ -56,14 +61,22 @@ func parseRequestBody[T any](r *http.Request) (*T, error) {
...
-// request body and if it's okay, will delegate to the provided callback with a
-// new HTTP request instance with the parse body in the context.
-func setRequestBodyInContext[T any](w http.ResponseWriter, r *http.Request, f func(w http.ResponseWriter, r *http.Request, body *T)) {
+// validateRequestBody is a helper method to avoid repetition. It parses
+// request body, validates it against the given validator instance and if it's
+// okay, will delegate to the provided callback with a new HTTP request instance
+// with the parse body in the context.
+//
+// Note that, technically, this method could be an ordinary (non-generic) method,
+// but it's defined as one to avoid confusion over value vs pointer arguments.
```
shipperizer, line 70: i'm a bit confused by this comment

alesstimec (approved):

rebac-admin-backend/v1/validation.go
```diff
@@ -56,14 +61,22 @@ func parseRequestBody[T any](r *http.Request) (*T, error) {
...
+func validateRequestBody[T any](v *validator.Validate, w http.ResponseWriter, r *http.Request, f func(w http.ResponseWriter, r *http.Request, body *T)) {
 	body, err := parseRequestBody[T](r)
 	if err != nil {
 		writeErrorResponse(w, err)
 		return
 	}
+	if err := v.Struct(body); err != nil {
+		writeErrorResponse(w, NewRequestBodyValidationError(err.Error()))
+		return
+	}
```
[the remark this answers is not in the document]
babakks: I removed generic usages.

rebac-admin-backend/v1/validation.go
babakks: This is not there anymore.

shipperizer (approved):

BarcoMasile (approved):

PR #245, opened 2024-03-25
IAM-773: drop non can_ relations from group entitlements

### LINKED ISSUE #243: Drop `assignee` relations from group entitlements endpoint
The endpoint `/api/v0/groups/{id}/entitlements` is performing a `ReadTuples` call to OpenFGA

we need to filter by anything that is a real "permissions" therefore anything starting with `can_`

shipperizer:
IAM-773: addressing group entitlements issue, closes #243

### DIFF
--- a/pkg/groups/service.go
+++ b/pkg/groups/service.go
@@ -382,6 +382,11 @@ func (s *Service) listPermissionsByType(ctx context.Context, ID, pType, continua
 	permissions := make([]string, 0)
 
 	for _, t := range r.GetTuples() {
+		// if relation doesn't start with can_ it means it's not a permission (see #assignee)
+		if !strings.HasPrefix(t.Key.Relation, "can_") {
+			continue
+		}
+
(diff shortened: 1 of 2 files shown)


pkg/groups/service.go
```diff
@@ -382,6 +382,11 @@ func (s *Service) listPermissionsByType(ctx context.Context, ID, pType, continua
 	permissions := make([]string, 0)
 
 	for _, t := range r.GetTuples() {
+		// if relation doesn't start with can_ it means it's not a permission (see #assignee)
+		if !strings.HasPrefix(t.Key.Relation, "can_") {
```
BarcoMasile, line 386: this will be needed for identities and roles also
next "tackle" should extract this into a common utilty/helper

BarcoMasile (approved):

pkg/groups/service.go
shipperizer: a role doesn't have any other relationship for now, a role cannot be `assigned` to be another `assignee` role

for identities, true, but not with the current model

PR #246, opened 2024-03-25
IAM 760

shipperizer:
IAM-760: empty id defaults to uuid on schemas and idps endpoints

- **fix: address empty IDs on schema and idp creation**
- **fix: introduce uri validation for params**

### DIFF
--- a/pkg/idp/handlers.go
+++ b/pkg/idp/handlers.go
@@ -176,15 +176,22 @@ func (a *API) handleCreate(w http.ResponseWriter, r *http.Request) {
 
 	}
 
+	if idp.ID != "" {
+		w.WriteHeader(http.StatusBadRequest)
+		json.NewEncoder(w).Encode(
+			types.Response{
+				Message: "IDP ID field is not allowed to be passed in",
+				Status:  http.StatusBadRequest,
+			},
+		)
+
+		return
+	}
+
 	idps, err := a.service.CreateResource(r.Context(), idp)
 
 	if err != nil {
 		status := http.StatusInternalServerError
-
-		if idps != nil {
-			status = http.StatusConflict
-		}
-
 		rr := types.Response{
 			Status:  status,
 			Message: err.Error(),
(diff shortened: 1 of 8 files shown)


pkg/idp/handlers.go
```diff
@@ -24,10 +24,10 @@ type API struct {
 
 func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	mux.Get("/api/v0/idps", a.handleList)
-	mux.Get("/api/v0/idps/{id}", a.handleDetail)
+	mux.Get("/api/v0/idps/{id:.*}", a.handleDetail)
 	mux.Post("/api/v0/idps", a.handleCreate)
-	mux.Patch("/api/v0/idps/{id}", a.handlePartialUpdate)
-	mux.Delete("/api/v0/idps/{id}", a.handleRemove)
+	mux.Patch("/api/v0/idps/{id:.*}", a.handlePartialUpdate)
+	mux.Delete("/api/v0/idps/{id:.*}", a.handleRemove)
```
BarcoMasile, line 30: nitpick: to make sure `id` is there, should be `{id:.+}`
same for the other similar change

pkg/idp/service.go
```diff
@@ -143,6 +143,12 @@ func (s *Service) CreateResource(ctx context.Context, data *Configuration) ([]*C
 	}
 
 	var idp *Configuration
+
+	// assign random ID if empty
+	if data.ID == "" {
+		data.ID = uuid.NewString()
+	}
```
BarcoMasile, line 150: issue: when creating new entitites, ID should never be user specified since it may conflict with backend-generated one (especially in the future). In case a user specified an ID, we should reject with bad request
Same goes for the other change

BarcoMasile (changes requested):
We need to avoid user-specified IDs for entities and reject with bad request all creation operations that provide an ID to use.

BarcoMasile (approved):

nsklikas (approved):

PR #247, opened 2024-03-26
feat: TokenPaginator enhancements + fix to properly handle empty tokens

### LINKED ISSUE #238: Fix pagination token generation to remove "old" pages
## Description
When settting tokens for new paginated response, we also need to delete pageTokens that don't have a following page, otherwise the last token for a page would always be present and the `PaginationHeader` method would never find the exiting condition to return `""`.

Example: `pkg/groups/handlers.go`

```golang
  for apiType, token := range pageTokens {
    paginator.SetToken(r.Context(), apiType, token)
  }
```

Old tokens would remain in the paginator and still end up in the base64 digest.

BarcoMasile:
## Description
This PR provides a new method to set multiple tokens at a time on the `TokenPaginator` object, while also adding tests for non-trivial methods.

## Changes
- sorted imports and added godocs
- added `SetTokens` method
  - now both `SetToken` and `SetTokens` methods don't actually set empty token values, in order to have an empty TokenHeader in case no continuation tokens are available
- added tests for non-trivial methods

## Fixes
#238

### DIFF
--- a/internal/http/types/token.go
+++ b/internal/http/types/token.go
@@ -9,9 +9,10 @@ import (
 	"encoding/json"
 	"net/http"
 
+	"go.opentelemetry.io/otel/trace"
+
 	"github.com/canonical/identity-platform-admin-ui/internal/logging"
 	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
-	"go.opentelemetry.io/otel/trace"
 )
 
 const (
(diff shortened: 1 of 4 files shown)


internal/http/types/token.go
```diff
@@ -51,15 +53,37 @@ func (p *TokenPaginator) LoadFromRequest(ctx context.Context, r *http.Request) e
...
+	if tokens == nil {
+		return
+	}
+
+	tokenMapCopy := make(map[string]string, len(tokens))
+	for key, value := range tokens {
+		if value != "" {
+			tokenMapCopy[key] = value
+		}
+	}
```
shipperizer, line 81: can we change it in place instead?

BarcoMasile: Yes! Done in new commit

shipperizer (approved):

shipperizer (approved):

PR #257, opened 2024-03-28
fix: export `GetRequestBodyFromContext` function

babakks:
This is to export the `GetRequestBodyFromContext` function.

Fixes CSS-7621


babakks:
Sorry, this is a wrong PR. We don't need to export the mentioned method.

PR #261, opened 2024-04-02
IAM-749: set openfga model and store env vars to dummy values for correct validation of admin-ui

shipperizer:
IAM-749: openfga client config validation needs store and model ids set, the k8s job will
substitute those with valid values in the dev setup


nsklikas (approved):

BarcoMasile (approved):
Maybe a hint in the readme would be beneficial

PR #265, opened 2024-04-03
feat: setup validation for APIs

BarcoMasile:
## Description
This PR introduces what's needed to have validation on each APIs, leveraging both custom `ValidationRegistry` struct (introduced by this PR) and 3rd party `validator` already part of the project dependencies.

## Changes
- updated license on touched files
- introduced `ValidationRegistry` object for registering validating functions for each API
- introduced 3rd party validator in each API struct to perform the actual validation
- tests on added code

### DIFF
--- a/internal/validation/registry_test.go
+++ b/internal/validation/registry_test.go
@@ -1,0 +1,188 @@
+// Copyright 2024 Canonical Ltd.
+// SPDX-License-Identifier: AGPL-3.0
+
+package validation
+
+import (
+	"context"
+	"net/http"
+	"net/http/httptest"
+	"testing"
+
+	"github.com/go-playground/validator/v10"
+	"go.opentelemetry.io/otel/trace"
+	"go.uber.org/mock/gomock"
+)
+
+//go:generate mockgen -build_flags=--mod=mod -package validation -destination ./mock_logger.go -source=../logging/interfaces.go
+//go:generate mockgen -build_flags=--mod=mod -package validation -destination ./mock_monitor.go -source=../monitoring/interfaces.go
+//go:generate mockgen -build_flags=--mod=mod -package validation -destination ./mock_tracing.go go.opentelemetry.io/otel/trace Tracer
+
+func TestValidator_Middleware(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	tracer := NewMockTracer(ctrl)
+	monitor := NewMockMonitorInterface(ctrl)
+	logger := NewMockLoggerInterface(ctrl)
+
+	tracer.EXPECT().
+		Start(gomock.Any(), gomock.Eq("validator.ValidationRegistry.ValidationMiddleware")).
+		Times(2).
+		Return(context.TODO(), trace.SpanFromContext(context.TODO()))
+
+	mainHandler := http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		_, _ = w.Write([]byte("main handler\n"))
+	})
+
+	vld := NewRegistry(tracer, monitor, logger)
+	vld.validatingFuncs["mock-key"] = func(r *http.Request) validator.ValidationErrors {
+		type InvalidStruct struct {
+			FirstName string `validate:"required"`
+		}
+
+		e := validator.New(validator.WithRequiredStructEnabled()).Struct(InvalidStruct{})
+		if e == nil {
+			return nil
+		}
+		return e.(validator.ValidationErrors)
+	}
+
+	for _, tt := range []struct {
+		name            string
+		expctedResponse string
+		expectedCode    int
+		mockRequest     *http.Request
+	}{
+		{
+			name:         "MatchingPrefix",
+			expectedCode: http.StatusBadRequest,
+			mockRequest:  httptest.NewRequest(http.MethodGet, "/api/v0/mock-key", nil),
+		},
+		{
+			name:            "NoMatchingPrefix",
+			expctedResponse: "main handler\n",
+			expectedCode:    http.StatusOK,
+			mockRequest:     httptest.NewRequest(http.MethodGet, "/api/v0/different-mock-key", nil),
+		},
+	} {
+		tt := tt
+		t.Run(tt.name, func(t *testing.T) {
+			mockResponse := httptest.NewRecorder()
+
+			decoratedHandler := vld.ValidationMiddleware(mainHandler)
+			decoratedHandler.ServeHTTP(mockResponse, tt.mockRequest)
+
+			responseValue := mockResponse.Body.String()
+
+			if mockResponse.Code != tt.expectedCode {
+				t.Fatalf("actual response code differes from expected")
+			}
+
+			if tt.expctedResponse == "" {
+				return
+			}
+
+			if responseValue != tt.expctedResponse {
+				t.Fatalf("actual response body differes from expected")
+			}
+		})
+	}
+}
+
+func TestValidator_RegisterValidator(t *testing.T) {
+	ctrl := gomock.NewController(t)
+	tracer := NewMockTracer(ctrl)
+	monitor := NewMockMonitorInterface(ctrl)
+	logger := NewMockLoggerInterface(ctrl)
+
+	emptyValidator := &ValidationRegistry{
+		validatingFuncs: make(map[string]ValidatingFunc),
+		tracer:          tracer,
+		monitor:         monitor,
+		logger:          logger,
+	}
+
+	noopVf := ValidatingFunc(func(r *http.Request) validator.ValidationErrors {
+		return nil
+	})
+	validatingFuncs := make(map[string]ValidatingFunc)
+	validatingFuncs["mock-key-1"] = noopVf
+
+	nonEmptyValidator := &ValidationRegistry{
+		validatingFuncs: validatingFuncs,
+		tracer:          tracer,
+		monitor:         monitor,
+		logger:          logger,
+	}
+
+	for _, tt := range []struct {
+		name      string
+		validator *ValidationRegistry
+		prefix    string
+		vf        ValidatingFunc
+		expected  string
+	}{
+		{
+			name:      "Nil middleware",
+			validator: emptyValidator,
+			prefix:    "",
+			vf:        nil,
+			expected:  "validatingFunc can't be null",
+		},
+		{
+			name:      "Existing key",
+			validator: nonEmptyValidator,
+			prefix:    "mock-key-1",
+			vf:        noopVf,
+			expected:  "key is already registered",
+		},
+		{
+			name:      "Success",
+			validator: emptyValidator,
+			prefix:    "mock-key",
+			vf:        noopVf,
+			expected:  "",
+		},
+	} {
+		tt := tt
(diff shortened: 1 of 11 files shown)


pkg/web/router.go
```diff
@@ -62,35 +65,44 @@ func NewRouter(idpConfig *idp.Config, schemasConfig *schemas.Config, rulesConfig
...
 	groups.NewAPI(
 		groups.NewService(externalConfig.OpenFGA(), wpool, tracer, monitor, logger),
 		tracer,
 		monitor,
 		logger,
-	).RegisterEndpoints(router)
+	).RegisterEndpoints(router).
+		SetupValidation(vldtr)
+
+	router = router.With(vldtr.Middleware).(*chi.Mux)
```
shipperizer, line 105: can we bundle this up with line 65 where authorization middleware is plugged in?

pkg/clients/handlers.go
```diff
@@ -220,7 +240,7 @@ func NewAPI(service ServiceInterface, logger logging.LoggerInterface) *API {
 	a := new(API)
 
 	a.service = service
-
+	a.validator = validator.New(validator.WithRequiredStructEnabled())
```
shipperizer, line 243: any benefit in having this passed in as a dependency and share it across the whole app?

internal/validation/types.go
```diff
@@ -0,0 +1,30 @@
...
+
+import (
+	"github.com/go-playground/validator/v10"
+)
+
+type ValidationError struct {
+	Status  int    `json:"status"`
+	Message string `json:"message"`
+	Data    []any  `json:"data,omitempty"`
+}
```
shipperizer, line 14: have a look at `internal/http/types` package, there is a response type in there

internal/validation/validator.go
```diff
@@ -0,0 +1,85 @@
...
+	tracer  tracing.TracingInterface
+	monitor monitoring.MonitorInterface
+	logger  logging.LoggerInterface
+}
+
+func (v *Validator) Middleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx, span := v.tracer.Start(r.Context(), "validator.Validator.Middleware")
+		r = r.WithContext(ctx)
+		defer span.End()
```
shipperizer, line 35:
```suggestion
		ctx, span := v.tracer.Start(r.Context(), "validator.Validator.Middleware")
		defer span.End()
		r = r.WithContext(ctx)
```

internal/validation/validator.go
```diff
@@ -0,0 +1,85 @@
...
+			if validationErr := vf(r); validationErr != nil {
+				e := NewValidationError(validationErr)
+
+				w.WriteHeader(e.Status)
+				_ = json.NewEncoder(w).Encode(e)
+				return
+			}
+		}
+
+		next.ServeHTTP(w, r)
```
shipperizer, line 50:
```suggestion
		if !isRegistered {
		   		next.ServeHTTP(w, r)
		}
		
		if validationErr := vf(r); validationErr != nil {
			e := NewValidationError(validationErr)
			w.WriteHeader(e.Status)
			json.NewEncoder(w).Encode(e)
			return
		}

		next.ServeHTTP(w, r)
```

internal/validation/validator.go
```diff
@@ -0,0 +1,85 @@
...
+}
+
+func (v *Validator) Middleware(next http.Handler) http.Handler {
+	return http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		ctx, span := v.tracer.Start(r.Context(), "validator.Validator.Middleware")
+		r = r.WithContext(ctx)
+		defer span.End()
+
+		prefix := v.prefix(r.URL.Path)
+		vf, isRegistered := v.validatingFuncs[prefix]
```
shipperizer, line 38:
```suggestion
		vf, ok := v.validatingFuncs[prefix]
```

internal/validation/validator.go
```diff
@@ -0,0 +1,85 @@
...
+		return fmt.Errorf("prefix is already registered")
+	}
+
+	v.validatingFuncs[prefix] = vf
+
+	return nil
+}
+
+func (v *Validator) prefix(path string) string {
+	after, found := strings.CutPrefix(path, fmt.Sprintf("/api/%s/", apiVersion))
```
shipperizer, line 69: as discussed, let's have a look again at the naming here, `path` doesn't sound correct

pkg/clients/handlers.go
```diff
@@ -29,12 +32,27 @@ type PaginationLinksResponse struct {
 	Next  string `json:"next,omitempty"`
 }
 
-func (a *API) RegisterEndpoints(mux *chi.Mux) {
+func (a *API) RegisterEndpoints(mux *chi.Mux) *API {
```
shipperizer, line 35:
```suggestion
func (a *API) RegisterEndpoints(mux *chi.Mux) {
```

as discussed, we don't really use the returning object
chaining pattern tends to be a pain to test, so if we can avoid it, better

pkg/clients/handlers.go
```diff
@@ -29,12 +32,27 @@ type PaginationLinksResponse struct {
...
 	mux.Get("/api/v0/clients", a.ListClients)
 	mux.Post("/api/v0/clients", a.CreateClient)
 	mux.Get("/api/v0/clients/{id}", a.GetClient)
 	mux.Put("/api/v0/clients/{id}", a.UpdateClient)
 	mux.Delete("/api/v0/clients/{id}", a.DeleteClient)
+
+	return a
+}
+
+func (a *API) SetupValidation(v *validation.Validator) *API {
```
shipperizer, line 45:
```suggestion
func (a *API) SetupValidation(v validation.ValidatorInterface) *API {
```

let's use interfaces, happy to use a shared one here instead of a local

pkg/groups/handlers.go
```diff
@@ -71,6 +75,21 @@ func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	mux.Get("/api/v0/groups/{id}/identities", a.handleListIdentities)
 	mux.Patch("/api/v0/groups/{id}/identities", a.handleAssignIdentities)
 	mux.Delete("/api/v0/groups/{id}/identities/{i_id}", a.handleRemoveIdentities)
+
+	return a
+}
+
+func (a *API) SetupValidation(v *validation.Validator) *API {
```
shipperizer, line 82:
```suggestion
func (a *API) RegisterValidation(v *validation.Validator) *API {
```

pkg/web/router.go
```diff
@@ -51,46 +54,56 @@ func NewRouter(idpConfig *idp.Config, schemasConfig *schemas.Config, rulesConfig
...
 
-	// apply authorization middleware using With to overcome issue with <id> URLParams not available
+	// apply authorization and validation middlewares using With to overcome issue with <id> URLParams not available
 	router = router.With(
 		authorization.NewMiddleware(
 			authorization.NewAuthorizer(externalConfig.Authorizer(), tracer, monitor, logger), monitor, logger).Authorize(),
+		vldtr.Middleware,
 	).(*chi.Mux)
 
+	router = router.With(vldtr.Middleware).(*chi.Mux)
```
shipperizer, line 64:
```suggestion
```

internal/validation/validator.go
```diff
@@ -0,0 +1,85 @@
...
+	"github.com/canonical/identity-platform-admin-ui/internal/logging"
+	"github.com/canonical/identity-platform-admin-ui/internal/monitoring"
+	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
+)
+
+const apiVersion = "v0"
+
+type ValidatingFunc func(r *http.Request) validator.ValidationErrors
+
+type Validator struct {
```
wood-push-melon, line 23: Just a non-blocking question. If I understand it correctly, the `Validator` here behaves like a registry of validation functions for handlers. Do you think it is better to name it with different names (e.g. `ValidationRegistry`)? Then this will not conflict with the naming of the validator in each handler's `API` struct.

pkg/clients/handlers.go
```diff
@@ -29,12 +32,27 @@ type PaginationLinksResponse struct {
...
+	err := v.RegisterValidatingFunc("clients", a.validatingFunc)
+	if err != nil {
+		a.logger.Fatal("unexpected validatingFunc already registered for clients")
+	}
+
+	return a
+}
+
+func (a *API) validatingFunc(r *http.Request) validator.ValidationErrors {
```
wood-push-melon, line 54: Just a non-blocking question. Is this where the `a.validator` will do the validation logic?

pkg/web/router.go
```diff
@@ -62,35 +65,44 @@ func NewRouter(idpConfig *idp.Config, schemasConfig *schemas.Config, rulesConfig
 	identities.NewAPI(
 		identities.NewService(externalConfig.KratosAdmin().IdentityApi(), tracer, monitor, logger),
 		logger,
-	).RegisterEndpoints(router)
+	).RegisterEndpoints(router).
+		SetupValidation(vldtr)
```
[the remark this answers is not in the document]
wood-push-melon: Similar question (non-blocking, just another thought). 

Is it OK to put the registration into the `NewAPI` factory method? The caller does not need to know how to instantiate the `API` object and other work in order to get a proper setup. This is not quite similar to the builder pattern.

The same question goes to `RegisterEndpoints`. We don't need to expose the `RegisterEndpoints` method to the outside I guess.

pkg/clients/handlers.go
wood-push-melon: > let's use interfaces, happy to use a shared one here instead of a local

Just a non-blocking question: why an interface is more appropriate than a struct here? Is it going to have different implementations of the validator registry?

pkg/clients/handlers.go
BarcoMasile: Yes, this is where the API struct will perform its validations leveraging the functionality from the validator object (3rd party)

internal/validation/validator.go
BarcoMasile: This is a great suggestion, although with recent changes this struct will actually centralize the validation process. I think I can still use your suggestion though, I like it! Thanks

pkg/web/router.go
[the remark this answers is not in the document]
BarcoMasile: @wood-push-melon that's what we'll eventually aim for

shipperizer (approved):

shipperizer:
@wood-push-melon completely valid point on the chaining/builder pattern, happy to create a ticket to refactor `NewApi`

internal/validation/registry_test.go
```diff
@@ -0,0 +1,180 @@
...
+	}{
+		{
+			name:         "MatchingPrefix",
+			expectedCode: 400,
+			mockRequest:  httptest.NewRequest(http.MethodGet, "/api/v0/mock-key", nil),
+		},
+		{
+			name:            "NoMatchingPrefix",
+			expctedResponse: "main handler\n",
+			expectedCode:    200,
```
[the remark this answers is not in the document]
BarcoMasile: missed these two bad boys, up in next rev

shipperizer (approved):

PR #273, opened 2024-04-12
IAM-769: add id validation to make sure it's never empty

### LINKED ISSUE #239: Add `chi` side validation for URLs + small fixes needed in `converters.go`
## Description
In `internal/authorization/converters.go`, L228, one thing we don't do here is actually check for the first id (`group_id`) to be non-blank, may raise issues at runtime if we don't validate endpoints path variables before this middleware is run, otherwise we would need to first check after parameter retrieval if the group_id is empty and the method allows for an empty parameter, like GET ( -> `can_view`) or POST ( -> `can_edit`).

Using regexp URL parameters to validate that selected endpoints have an ID path variable populated is the best way to fix this.
From the doc:
```golang
  // Regexp url parameters:
  r.Get("/{articleSlug:[a-z-]+}", getArticleBySlug) // GET /articles/home-is-toronto
```

#### Bonus nitpick
`/entitlements` suffixed endpoints for group don't support POST operations in this PR, only GET, PATCH and DELETE

shipperizer:
IAM-769: closes #239

### DIFF
--- a/pkg/clients/handlers.go
+++ b/pkg/clients/handlers.go
@@ -35,9 +35,9 @@ type PaginationLinksResponse struct {
 func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	mux.Get("/api/v0/clients", a.ListClients)
 	mux.Post("/api/v0/clients", a.CreateClient)
-	mux.Get("/api/v0/clients/{id}", a.GetClient)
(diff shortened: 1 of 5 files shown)


BarcoMasile:
Done in https://github.com/canonical/identity-platform-admin-ui/pull/275

BarcoMasile (approved):

PR #274, opened 2024-04-12
UI: add pagination and identity creation

edlerd:
# Done
- Add pagination to clients, schemas and identity lists in ui.
- Add identity creation form
- Add identity deletion

fixes WD-10253

Screenshot from 2024-04-17 19-49-51
Screenshot from 2024-04-17 19-49-45

### DIFF
--- a/ui/src/components/Pagination.tsx
+++ b/ui/src/components/Pagination.tsx
@@ -1,0 +1,47 @@
+import React, { FC } from "react";
+import { PaginatedResponse } from "types/api";
+import { Button } from "@canonical/react-components";
+import { usePagination } from "util/usePagination";
+import IconLeft from "components/IconLeft";
+import IconRight from "components/IconRight";
+
+interface Props {
+  response?: PaginatedResponse<unknown[]>;
+}
+
+const Pagination: FC<Props> = ({ response }) => {
+  const { pageToken, setPageToken } = usePagination();
+  const showFirstLink = pageToken !== "";
+  const isMissingNext = !response || !response._meta.next;
+  const isEmptyPage = response?.data.length === 0;
+
+  if (!showFirstLink && (isMissingNext || isEmptyPage)) {
+    return null;
+  }
+
+  const next = response?._meta.next ?? "";
+  const prev = response?._meta.prev ?? "";
+
+  return (
+    <>
+      {showFirstLink && (
+        <Button onClick={() => setPageToken("")} title="First page">
+          <IconLeft />
+          <IconLeft />
+        </Button>
+      )}
+      {prev && (
+        <Button onClick={() => setPageToken(prev)} title="Previous page">
+          <IconLeft />
+        </Button>
+      )}
+      {next && (
+        <Button onClick={() => setPageToken(next)} title="Next page">
+          <IconRight />
+        </Button>
+      )}
+    </>
+  );
+};
+
+export default Pagination;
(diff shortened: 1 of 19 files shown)


ui/src/components/Pagination.tsx
```diff
@@ -0,0 +1,60 @@
...
+  const last = response?._meta.last ?? "";
+  const next = response?._meta.next ?? "";
+  const prev = response?._meta.prev ?? "";
+
+  const leftIcon = (
+    <i className="p-icon--chevron-down" style={{ rotate: "90deg" }} />
+  );
+  const rightIcon = (
+    <i className="p-icon--chevron-down" style={{ rotate: "270deg" }} />
+  );
```
BarcoMasile, line 30: nitpick(non-blocking): these two would probably benefit from being defined outside of the component

BarcoMasile (dismissed):

ui/src/pages/clients/ClientList.tsx
```diff
@@ -8,13 +8,17 @@ import { isoTimeToString } from "util/date";
 import usePanelParams from "util/usePanelParams";
 import EditClientBtn from "pages/clients/EditClientBtn";
 import DeleteClientBtn from "pages/clients/DeleteClientBtn";
+import Loader from "components/Loader";
+import Pagination from "components/Pagination";
+import { usePagination } from "util/usePagination";
```
BarcoMasile, line 13: possible issue(?): client list after recent changes relies on `page_token` instead of the `page` parameter. If I'm not mistaken, this refers to the `page` parameter instead.

edlerd: I tested this with latest main branch and the parameter names now seem to be aligned, so we always use `page_token`. At least it worked for me locally. @shipperizer right?

ui/src/components/Pagination.tsx
edlerd: Extracted those two.

ui/src/types/api.d.ts
```diff
@@ -4,6 +4,15 @@ export interface ApiResponse<T> {
   status: number;
 }
 
+export type PaginatedResponse<T> = {
+  _meta: {
+    first?: string;
+    last?: string;
```
shipperizer, line 10: can probably drop these

ui/src/util/usePagination.tsx
```diff
@@ -0,0 +1,11 @@
+import { useSearchParams } from "react-router-dom";
+
+export const usePagination = () => {
+  const [searchParams, setSearchParams] = useSearchParams();
+  const page = searchParams.get("page") ?? "";
```
shipperizer, line 5: page_token

ui/src/util/usePagination.tsx
```diff
@@ -0,0 +1,11 @@
...
+  const setPage = (newPage: string) => {
+    setSearchParams({ ...searchParams, page: newPage });
```
shipperizer, line 7: page_token

ui/src/pages/clients/ClientList.tsx
shipperizer: yes page_token is the one to use

not a good react reviewer but it seems you are still using page instead of page_token

this is what @BarcoMasile is pointing out I guess

ui/src/util/usePagination.tsx
edlerd: This (and the others below are the parameter names used by the frontend, exposed to the user. But I guess we can keep them consistent with the backend.

ui/src/pages/clients/ClientList.tsx
edlerd: The page was only used in the browser, for the api request we'd use page_token. I changed that now, so the browser also uses page_token, so it is more consistent.

shipperizer (approved):

PR #275, opened 2024-04-12
IAM 776 Implement validation for `groups`, `schemas`, `identities` handlers

BarcoMasile:
## Description
This PR introduces the first implementation for the validation layer. Validation is delegated to each validatingFunction that handlers register through the `RegisterValidation` method. The process is run inside one middleware provided by the `ValidationRegistry` object.

Example:
```golang
  groupsAPI := groups.NewAPI(/*...*/)
  groupsAPI.RegisterValidation(vldtr)
```

## Changes
- enahnced Validation errors with detailed errors for the UI to consume
- added tests
- api struct validators (3rd party) enhanced to let errors use the json field name to avoid discrepancy between fields used by UI and fields referenced by backend errors
- added validation constraint on URL Params for endpoints to forbid empty values
- added implementation of validation process for the following handlers
  - schemas
  - groups
  - identities
 

### Missing
Tests for new validation functions for handlers.

### DIFF
--- a/internal/validation/registry.go
+++ b/internal/validation/registry.go
@@ -4,13 +4,17 @@
 package validation
 
 import (
+	"bytes"
+	"context"
 	"encoding/json"
 	"fmt"
+	"io"
 	"net/http"
 	"strings"
 
 	"github.com/go-playground/validator/v10"
 
+	"github.com/canonical/identity-platform-admin-ui/internal/http/types"
 	"github.com/canonical/identity-platform-admin-ui/internal/logging"
 	"github.com/canonical/identity-platform-admin-ui/internal/monitoring"
 	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
@@ -20,13 +24,16 @@ const apiVersion = "v0"
 
 type ValidationRegistryInterface interface {
 	ValidationMiddleware(next http.Handler) http.Handler
-	RegisterValidatingFunc(key string, vf ValidatingFunc) error
+	RegisterPayloadValidator(key string, vf PayloadValidatorInterface) error
 }
 
-type ValidatingFunc func(r *http.Request) validator.ValidationErrors
+type PayloadValidatorInterface interface {
+	NeedsValidation(r *http.Request) bool
+	Validate(ctx context.Context, method, endpoint string, body []byte) (context.Context, validator.ValidationErrors, error)
+}
 
 type ValidationRegistry struct {
-	validatingFuncs map[string]ValidatingFunc
+	validators map[string]PayloadValidatorInterface
 
 	tracer  tracing.TracingInterface
 	monitor monitoring.MonitorInterface
@@ -41,34 +48,69 @@ func (v *ValidationRegistry) ValidationMiddleware(next http.Handler) http.Handle
 		r = r.WithContext(ctx)
 		key := v.apiKey(r.URL.Path)
 
-		vf, ok := v.validatingFuncs[key]
-		if !ok {
+		payloadValidator, ok := v.validators[key]
+		if !ok || !payloadValidator.NeedsValidation(r) {
 			next.ServeHTTP(w, r)
 			return
 		}
 
-		if validationErr := vf(r); validationErr != nil {
-			e := NewValidationError(validationErr)
+		reqBody := r.Body
+		defer reqBody.Close()
+		body, err := io.ReadAll(reqBody)
+
+		if err != nil {
+			badRequestFromError(w, NoBodyError)
+			return
+		}
+
+		// don't break existing handlers, replace the body that was consumed
+		r.Body = io.NopCloser(bytes.NewReader(body))
+
+		endpoint, _ := ApiEndpoint(r.URL.Path, key)
+		var validationErr validator.ValidationErrors
+
+		ctx, validationErr, err = payloadValidator.Validate(r.Context(), r.Method, endpoint, body)
+
+		if err != nil {
+			badRequestFromError(w, err)
+			return
+		}
+
+		// handler validation errors
+		if validationErr != nil {
+			e := NewValidationError("validation errors", validationErr)
 
 			w.WriteHeader(e.Status)
 			_ = json.NewEncoder(w).Encode(e)
 			return
 		}
 
-		next.ServeHTTP(w, r)
+		// if no errors, proceed with the request
+		next.ServeHTTP(w, r.WithContext(ctx))
 	})
 }
 
-func (v *ValidationRegistry) RegisterValidatingFunc(key string, vf ValidatingFunc) error {
-	if vf == nil {
-		return fmt.Errorf("validatingFunc can't be null")
+func badRequestFromError(w http.ResponseWriter, err error) {
+	w.WriteHeader(http.StatusBadRequest)
+	_ = json.NewEncoder(w).Encode(
+		types.Response{
+			Message: err.Error(),
+			Status:  http.StatusBadRequest,
+		},
+	)
+	return
+}
+
+func (v *ValidationRegistry) RegisterPayloadValidator(key string, payloadValidator PayloadValidatorInterface) error {
+	if payloadValidator == nil {
+		return fmt.Errorf("payloadValidator can't be null")
 	}
 
-	if _, ok := v.validatingFuncs[key]; ok {
+	if _, ok := v.validators[key]; ok {
 		return fmt.Errorf("key is already registered")
 	}
 
-	v.validatingFuncs[key] = vf
+	v.validators[key] = payloadValidator
 
 	return nil
 }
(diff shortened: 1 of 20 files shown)


internal/validation/registry_test.go
```diff
@@ -88,6 +86,26 @@ func TestValidator_Middleware(t *testing.T) {
...
+		if name == "-" {
+			return ""
+		}
+		return name
+	})
+	e := validate.Struct(InvalidStruct{})
+	if e == nil {
+		return nil
+	}
+	return e.(validator.ValidationErrors)
```
shipperizer, line 106: why not simply creating the struct with a message?is the error more composite?

internal/validation/types.go
```diff
@@ -4,21 +4,66 @@
...
+		)
+		failedValidations[field] = failures
+	}
+
+	return failedValidations
+}
+
+func NewValidator() *validator.Validate {
+	validate := validator.New(validator.WithRequiredStructEnabled())
+	validate.RegisterTagNameFunc(func(fld reflect.StructField) string {
```
shipperizer, line 60: add a comment to explain what is this for (the func u register)

pkg/groups/handlers.go
```diff
@@ -25,10 +26,11 @@ import (
 const (
 	ROLE_TOKEN_KEY  = "roles"
 	GROUP_TOKEN_KEY = "groups"
+	API_KEY         = "groups"
 )
 
 type UpdateRolesRequest struct {
-	Roles []string `json:"roles" validate:"required"`
+	Roles []string `json:"roles" validate:"required,dive,required"`
```
shipperizer, line 33:
```suggestion
	Roles []string `json:"roles" validate:"required,dive"`
```

double required?or is it by design?

pkg/groups/validation.go
```diff
@@ -0,0 +1,33 @@
...
+	return r.Method == http.MethodPost && strings.HasSuffix(endpoint, "/roles")
+}
+
+func isAssignPermissions(r *http.Request, endpoint string) bool {
+	return r.Method == http.MethodPatch && strings.HasSuffix(endpoint, "/entitlements")
+}
+
+func isAssignIdentities(r *http.Request, endpoint string) bool {
+	return r.Method == http.MethodPatch && strings.HasSuffix(endpoint, "/identities")
+}
```
shipperizer, line 33: as much as this is stateless, I would still make it into a struct

pkg/groups/handlers.go
```diff
@@ -25,10 +26,11 @@ import (
 const (
 	ROLE_TOKEN_KEY  = "roles"
 	GROUP_TOKEN_KEY = "groups"
+	API_KEY         = "groups"
```
shipperizer, line 29: feels like an API attribute now

pkg/groups/handlers.go
```diff
@@ -62,30 +64,102 @@ type API struct {
...
+	if isAssignIdentities(r, endpoint) {
+		updateIdentities := new(UpdateIdentitiesRequest)
+		if err := json.Unmarshal(body, updateIdentities); err != nil {
+			return nil, err
+		}
+
+		err = a.validator.Struct(updateIdentities)
+		validated = true
+	}
+
```
shipperizer, line 153: this is more a switch I d say

pkg/groups/handlers.go
```diff
@@ -62,30 +64,102 @@ type API struct {
...
+	err := v.RegisterValidatingFunc(API_KEY, a.validatingFunc)
 	if err != nil {
 		a.logger.Fatal("unexpected validatingFunc already registered for groups")
 	}
 }
 
-func (a *API) validatingFunc(r *http.Request) validator.ValidationErrors {
-	return nil
+func (a *API) validatingFunc(r *http.Request) (validator.ValidationErrors, error) {
+	if !shouldValidate(r) {
```
shipperizer, line 90: I ll throw it out there as it seems a good idea but not sure it s fine to implement 

API cares that a validation happens, but doesn't care about the validation rules themselves 

so what if we offload the content of this function to another interface that deals with it?

Chinese boxing pattern, I know, but it feels like this should be somewhere else and not in the API

pkg/identities/handlers.go
```diff
@@ -36,24 +41,81 @@ type API struct {
...
 	if err != nil {
 		a.logger.Fatal("unexpected validatingFunc already registered for identities")
 	}
+
+	a.setupValidator()
 }
 
-func (a *API) validatingFunc(r *http.Request) validator.ValidationErrors {
-	return nil
+func (a *API) setupValidator() {
```
shipperizer, line 62: I'm missing where this is called

internal/validation/registry_test.go
BarcoMasile: yes it is
but I wanted "real errors" since I need to test the function with a proper object

pkg/groups/handlers.go
BarcoMasile: Great idea, I'll proceed

pkg/identities/handlers.go
BarcoMasile: L59

pkg/clients/handlers.go
```diff
@@ -10,39 +10,35 @@ import (
...
 
 func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	mux.Get("/api/v0/clients", a.ListClients)
 	mux.Post("/api/v0/clients", a.CreateClient)
-	mux.Get("/api/v0/clients/{id:.+}", a.GetClient)
-	mux.Put("/api/v0/clients/{id:.+}", a.UpdateClient)
-	mux.Delete("/api/v0/clients/{id:.+}", a.DeleteClient)
+	mux.Get("/api/v0/clients/{id}", a.GetClient)
+	mux.Put("/api/v0/clients/{id}", a.UpdateClient)
+	mux.Delete("/api/v0/clients/{id}", a.DeleteClient)
```
shipperizer, line 32:
```suggestion
	mux.Get("/api/v0/clients/{id:.+}", a.GetClient)
	mux.Put("/api/v0/clients/{id:.+}", a.UpdateClient)
	mux.Delete("/api/v0/clients/{id:.+}", a.DeleteClient)
```

rebase went wrong here

pkg/clients/handlers.go
```diff
@@ -203,9 +199,10 @@ func (a *API) parseListClientsRequest(r *http.Request) (*ListClientsRequest, err
 
 func NewAPI(service ServiceInterface, logger logging.LoggerInterface) *API {
 	a := new(API)
+	a.apiKey = "clients"
 
 	a.service = service
-	a.validator = validator.New(validator.WithRequiredStructEnabled())
+	//a.payloadValidator = NewClientsPayloadValidator(a.apiKey)
```
shipperizer, line 205: is it missing on purpose?

pkg/groups/handlers.go
```diff
@@ -28,7 +26,7 @@ const (
 )
 
 type UpdateRolesRequest struct {
-	Roles []string `json:"roles" validate:"required"`
+	Roles []string `json:"roles" validate:"required,dive,required"`
```
shipperizer, line 29: please add a comment saying the double required is needed by `dive`, just until we get familiar, a link to the docs is as good as a sentence happy with anything

pkg/groups/validation.go
```diff
@@ -0,0 +1,112 @@
...
+
+	"github.com/canonical/identity-platform-admin-ui/internal/validation"
+)
+
+type PayloadValidator struct {
+	apiKey    string
+	validator *validator.Validate
+}
+
+func (_ *PayloadValidator) NeedsValidation(r *http.Request) bool {
```
shipperizer, line 22:
```suggestion
func (p *PayloadValidator) NeedsValidation(r *http.Request) bool {
```
unless linter complains I d leave the p, easier to understand it belongs (mainly a thing for grepping)

pkg/clients/handlers.go
BarcoMasile: Yes, placeholder for next task

pkg/clients/handlers.go
BarcoMasile: Most likely :D

internal/validation/types.go
```diff
@@ -4,21 +4,69 @@
...
 	"github.com/canonical/identity-platform-admin-ui/internal/http/types"
 )
 
-func NewValidationError(errors validator.ValidationErrors) *types.Response {
+var (
+	NoBodyError = errors.New("request body is not present")
+)
+
+func NoMatchError(apiKey string) error {
+	return fmt.Errorf("can't find matching validation process for '%s' endpoint", apiKey)
```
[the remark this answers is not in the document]
BarcoMasile: You're right, this started as an exported field but it's not needed (as exported) anymore, I'll change it in next PR

shipperizer (approved):

internal/validation/registry.go
```diff
@@ -41,34 +48,69 @@ func (v *ValidationRegistry) ValidationMiddleware(next http.Handler) http.Handle
...
+
+		if err != nil {
+			badRequestFromError(w, NoBodyError)
+			return
+		}
+
+		// don't break existing handlers, replace the body that was consumed
+		r.Body = io.NopCloser(bytes.NewReader(body))
+
+		endpoint, _ := ApiEndpoint(r.URL.Path, key)
```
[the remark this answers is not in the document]
BarcoMasile: will try to find a better one, couldn't find one still

Issue #228, opened 2024-03-19
Using go-playground/validate for payload validation 


shipperizer:
we noticed that some of the endpoints are failing to do proper payload validation, we do have some annotations on structs but we failed to get a proper setup 

we should pivot to use https://github.com/go-playground/validator and enforce required parameters on JSON payload, see example in  https://github.com/go-playground/validator/blob/master/_examples%2Fsimple%2Fmain.go

this issue relates to #227

PR #279, opened 2024-04-18
IAM 776 Implement validation for all handlers

BarcoMasile:
## Description
This PR implements validation for remaining APIs, and add tests for all of them.
It also improves previously added tests from #275 .
It also enables validation (basically reverting the changes from #278 ), while also renaming the variable to something more explicit.
Added comments on validation tags to be explicit about what validation is performed.

### DIFF
--- a/pkg/clients/handlers.go
+++ b/pkg/clients/handlers.go
@@ -202,7 +202,7 @@ func NewAPI(service ServiceInterface, logger logging.LoggerInterface) *API {
 	a.apiKey = "clients"
 
 	a.service = service
-	//a.payloadValidator = NewClientsPayloadValidator(a.apiKey)
+	a.payloadValidator = NewClientsPayloadValidator(a.apiKey)
 	a.logger = logger
 
 	return a
--- a/pkg/clients/handlers.go
+++ b/pkg/clients/handlers.go
@@ -35,7 +35,7 @@ func (a *API) RegisterEndpoints(mux *chi.Mux) {
 func (a *API) RegisterValidation(v validation.ValidationRegistryInterface) {
 	err := v.RegisterPayloadValidator(a.apiKey, a.payloadValidator)
 	if err != nil {
-		a.logger.Fatalf("unexpected validatingFunc already registered for clients, %s", err)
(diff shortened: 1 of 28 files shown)


pkg/clients/validation.go
```diff
@@ -0,0 +1,87 @@
...
+
+	if !validated {
+		return ctx, nil, validation.NoMatchError(p.apiKey)
+	}
+
+	if err == nil {
+		return ctx, nil, nil
+	}
+
+	return ctx, err.(validator.ValidationErrors), nil
```
shipperizer, line 68: context is not touched here, is the plan to add the unmarshaled body to it in a separate pr or is it a miss?

pkg/clients/handlers.go
```diff
@@ -27,15 +27,15 @@ type API struct {
...
+	mux.Get("/api/v0/clients/{id:.+}", a.GetClient)
+	mux.Put("/api/v0/clients/{id:.+}", a.UpdateClient)
+	mux.Delete("/api/v0/clients/{id:.+}", a.DeleteClient)
 }
 
 func (a *API) RegisterValidation(v validation.ValidationRegistryInterface) {
 	err := v.RegisterPayloadValidator(a.apiKey, a.payloadValidator)
 	if err != nil {
-		a.logger.Fatalf("unexpected validatingFunc already registered for clients, %s", err)
+		a.logger.Fatalf("unexpected error while registering PayloadValidator for 'clients', %s", err)
```
shipperizer, line 38:
```suggestion
		a.logger.Fatalf("unexpected error while registering PayloadValidator for clients, %s", err)
```

shipperizer (approved):

pkg/clients/validation.go
BarcoMasile: Yes, this will be part of the refactoring of payload handling and it will be done in the near future to avoid "double" deserialization

pkg/clients/handlers.go
BarcoMasile: I'll do it for all occurrences and re-push

shipperizer (approved):

shipperizer:
closes #228

PR #295, opened 2024-04-29
IAM 855 fix roles handlers

### LINKED ISSUE #294: Can't remove roles or identities from a group
When doing a DELETE to /groups/group1/roles/role1 or /groups/group1/identities/identity1 to remove the role or identity from a group I get a 403 response with this error:

```
{"data":null,"message":"insufficient permissions to execute operation","status":403,"_meta":null}
```

I'm on latest main (commit: 0cc2bcece295d2014a269ab5c78154aeac71b383).

I have reset my tuples to those from the [seeding doc](https://github.com/canonical/identity-platform-admin-ui/wiki/OpenFGA-setup#seeding) and appear to have delete permissions (I'm passing the token johndoe as the auth header):

```
$ fga query check --store-id=$STORE_ID user:johndoe can_delete identity:global --model-id=$MODEL_ID
{
  "allowed":true,
  "resolution":""
}
$ fga query check --store-id=$STORE_ID user:johndoe can_delete role:global --model-id=$MODEL_ID
{
  "allowed":true,
  "resolution":""
}
```

I can DELETE entitlements from the group (/groups/group1/entitlements/entitlement1).

BarcoMasile:
## Description
This PR fixes a couple of unexpected behavior when invoking the role API.
Fixes #294 .

## Changes
- CreateRole now adds the explicit entitlement `can_view` to `user:*` since it would not work otherwise
- GetRoleDetail now returns
  - the role id in case it's present and the user is allowed to view it
  - 404 in case the role is not present and the user is allowed to view
  - 403 in case the user is **not** allowed to view it (no matter if the role is present or not, this info is concealed from the unauthorized user)
- DeleteRole now removes also all direct associations from the role, meaning all `can_view`, `can_edit`, `can_delete`, `can_create`, `assignee`, `privileged` are removed when the role is removed
- tests are adjusted

NB: a subsequent PR will take care of returning errors from the concurrent deletions, this requires more work on the worker pool side and on the role service, so it needs to be done separately. This is way the remove role test is not testing error for now.


pkg/roles/handlers.go
```diff
@@ -146,6 +146,17 @@ func (a *API) handleDetail(w http.ResponseWriter, r *http.Request) {
 		return
 	}
 
+	if role == "" {
+		w.WriteHeader(http.StatusNotFound)
+		json.NewEncoder(w).Encode(
+			types.Response{
+				Message: "Rule not found",
```
shipperizer, line 153:
```suggestion
				Message: "Role not found",
```

pkg/roles/service.go
```diff
@@ -112,18 +114,33 @@ func (s *Service) CreateRole(ctx context.Context, userID, ID string) error {
...
+		s.logger.Error(err.Error())
+		return err
+	}
+
+	// after successful role creation, `can_view` to user:* needs to be assigned
+	// explicitly for it to work
+	err = s.ofga.WriteTuples(
+		ctx,
+		*ofga.NewTuple(ALL_USERS, CAN_VIEW_RELATION, role),
 	)
```
shipperizer, line 144: I would say we should bundle all the writes in one go on the above, chances that fails but this doesn't are quite small unless we have a conflict which we should check somehow beforehand

pkg/roles/service_test.go
```diff
@@ -418,13 +427,37 @@ func TestServiceCreateRole(t *testing.T) {
...
+							t.Errorf("expected tuples to be %v got %v", ps, tuples)
+						}
+
+						return test.secondExpected
+					},
+				)
+			}
+
 			if test.expected != nil {
 				mockLogger.EXPECT().Error(gomock.Any()).Times(1)
```
shipperizer, line 447: try mintimes and maxtimes to remove the secondExpected

shipperizer: ok seemsore complex than I thought, worth a chat

pkg/roles/service_test.go
```diff
@@ -585,18 +621,19 @@ func TestServiceDeleteRole(t *testing.T) {
 			} else {
 				// TODO @shipperizer fix this so that we can pin it down to the error case only
 				mockLogger.EXPECT().Errorf(gomock.Any(), gomock.Any(), gomock.Any()).AnyTimes()
-				mockOpenFGA.EXPECT().DeleteTuples(
+				/*mockOpenFGA.EXPECT().DeleteTuples(
```
shipperizer, line 624: drop comments when done

pkg/roles/service_test.go
BarcoMasile: Bundling all the writes in one go as you suggested will remove the need for a secondExpected

shipperizer (approved):

PR #297, opened 2024-04-29
IAM 834 Use `name` identifier for `Groups`, and `Roles` creation

### LINKED ISSUE #271: Let users specify `name` instead of `ID` for Groups and Roles
In some cases the field needs to be added. UI team needs to be alerted about this so they can uniform the use of the API.
Plus the following resources suffer from "ID overwrite" in case of update:
- Identities
- IdPs
  
  
For the Rules the ID check is performed in the service layer instead of the controller.

BarcoMasile:
## Description
This PR tries to bring uniformity to Admin UI APIs in order to avoid breaking changes in the future (or at least reduce them to the least possible). This involves never letting users specify ID for resource creations, but rely on a different identifier.
Also, now invocations like `GET /groups/{id}` will return a json instead of a single string. The json response will include both the ID and the identifier provided at creation time.

This fixes #271

### DIFF
--- a/pkg/groups/handlers.go
+++ b/pkg/groups/handlers.go
@@ -40,8 +40,9 @@ type UpdatePermissionsRequest struct {
 	Permissions []Permission `json:"permissions" validate:"required,dive,required"`
 }
 
-type GroupRequest struct {
-	ID string `json:"id" validate:"required"`
+type Group struct {
+	ID   string `json:"id,omitempty"`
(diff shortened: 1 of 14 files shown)


shipperizer (approved):

shipperizer:
@BarcoMasile u need at least a `fix` commit to trigger a new release 

if we want o release this rewrite the commit history to get at least one of those a `fix`

PR #300, opened 2024-04-30
IAM 867 fix groups handlers

BarcoMasile:
## Description
This PR fixes unexpected behavior when invoking the group API.

Similar to the work done in #295 .

## Changes
- CreateGroup now adds the explicit entitlement `can_view` to user:{userID} since it would not work otherwise
- GetGroupDetail now returns
  - the group id in case it's present and the user is allowed to view it
  - 404 in case the group is not present and the user is allowed to view
  - 403 in case the user is not allowed to view it (no matter if the role is present or not, this info is concealed from the unauthorized user)
- DeleteRole now removes also all direct associations from the group, meaning all `can_view`, `can_edit`, `can_delete`, `can_create`, `member`, `privileged` are removed when the group is removed
- tests are adjusted

NB: a subsequent PR will take care of returning errors from the concurrent deletions, this requires more work on the worker pool side and on the role service, so it needs to be done separately. This is way the remove role test is not testing error for now.

### DIFF
--- a/pkg/groups/service.go
+++ b/pkg/groups/service.go
@@ -21,6 +21,8 @@ import (
 const (
 	MEMBER_RELATION   = "member"
 	ASSIGNEE_RELATION = "assignee"
+	// TODO: @barco centralize common relation name definitions
+	CAN_VIEW_RELATION = "can_view"
 )
 
 type listPermissionsResult struct {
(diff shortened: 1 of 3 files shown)


nsklikas (dismissed):

pkg/groups/service.go
```diff
@@ -458,7 +470,42 @@ func (s *Service) removePermissionsByType(ctx context.Context, ID, pType string)
...
+	if len(directs) == 0 {
+		return
+	}
+
+	if err := s.ofga.DeleteTuples(ctx, directs...); err != nil {
+		s.logger.Error(err.Error())
+	}
+}
+
+/*func (s *Service) removeMembers(ctx context.Context, ID string) {
```
shipperizer, line 508: drop comments

BarcoMasile: Thanks

nsklikas (approved):

PR #310, opened 2024-05-24
IAM 829 Use token pagination strategy for Rules API

### LINKED ISSUE #268: Standardize rules API pagination strategy
at the moment the `/api/v0/rules` API follows the `page|size` pagination strategy 

this goes against what is being used in 

- `api/v0/identities`
- `api/v0/clients`
- `api/v0/schemas`

plan is to standardzie around a `page_token` strategy like the APIs above, using a base64 json representation of an offset:

```json
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "type": "object",
  "properties": {
    "offset": {
      "type": "string"
    }
  },
  "required": [
    "offset"
  ]
}
```

```bash
shipperizer in ~/shipperizer/identity-platform-admin-ui/pkg/rules on IAM-782 ● λ echo -n '{"offset":"250"}' | base64 | base64 -d
{"offset":"250"}%                                                                                                                           
shipperizer in ~/shipperizer/identity-platform-admin-ui/pkg/rules on IAM-782 ● λ echo -n '{"offset":"250"}' | base64            
eyJvZmZzZXQiOiIyNTAifQ==

```

this is the same token schema used by the `/api/v0/schemas`

once we get the offset and the page size, we should be more than able to interpolate the page

BarcoMasile:
## Description
This PR adopts the page_token strategy for pagination of Rules List API, with an encoded offset.

Fixes #268

### DIFF
--- a/pkg/rules/handlers.go
+++ b/pkg/rules/handlers.go
@@ -9,7 +9,6 @@ import (
 	"fmt"
 	"io"
 	"net/http"
-	"strconv"
 
 	"github.com/canonical/identity-platform-admin-ui/internal/http/types"
 	"github.com/canonical/identity-platform-admin-ui/internal/logging"
@@ -19,7 +18,7 @@ import (
 	oathkeeper "github.com/ory/oathkeeper-client-go"
 )
 
-const DEFAULT_PAGE_NUMBER = 1
+const DEFAULT_OFFSET int64 = 0
 
 type API struct {
 	apiKey           string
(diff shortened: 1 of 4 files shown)


shipperizer (approved):

pkg/rules/handlers.go
```diff
@@ -59,21 +64,71 @@ func (a *API) handleList(w http.ResponseWriter, r *http.Request) {
 		}
 
 		w.WriteHeader(http.StatusInternalServerError)
-		json.NewEncoder(w).Encode(rr)
+		_ = json.NewEncoder(w).Encode(rr)
```
nsklikas, line 67: Is this for the linter?

nsklikas (approved):

pkg/rules/handlers.go
BarcoMasile: Luckily the linter doesn't complain (my IDE does) but I prefer doing it to explicitly mark that error as ignored, since in case of an error there's no other thing that function can do

PR #313, opened 2024-06-05
Add UI serving

nsklikas:
- Starts serving the ui pages under the `/ui` path
- The ui uses relative paths to fetch its assets

### DIFF
--- a/pkg/ui/handlers.go
+++ b/pkg/ui/handlers.go
@@ -1,9 +1,12 @@
+// Copyright 2024 Canonical Ltd.
+// SPDX-License-Identifier: AGPL-3.0
+
 package ui
 
 import (
 	"io/fs"
 	"net/http"
-	"path"
+	"strings"
 
 	"github.com/canonical/identity-platform-admin-ui/internal/logging"
 	"github.com/canonical/identity-platform-admin-ui/internal/monitoring"
@@ -24,15 +27,19 @@ type Config struct {
 }
 
 func (a *API) RegisterEndpoints(mux *chi.Mux) {
-	// Add the trailing slash if it's missing
 	mux.NotFound(a.uiFiles)
 }
 
-// TODO: Validate response when server error handling is implemented
 func (a *API) uiFiles(w http.ResponseWriter, r *http.Request) {
-	// If the path is not /ui/ add the html suffix to make file system serving work
-	if ext := path.Ext(r.URL.Path); ext == "" && r.URL.Path != "/" {
-		r.URL.Path += "%s.html"
+	// If a `/api` request ends up here, it means that the route does not exist
+	// Do not rely on the UI to return the 404 response
+	if strings.HasPrefix(r.URL.Path, "/api") {
+		http.NotFound(w, r)
+		return
+	}
+	// This is a SPA, everything HTML page uses the same `index.html`
+	if !strings.HasPrefix(r.URL.Path, "/assets") {
+		r.URL.Path = "/"
 	}
 
 	// Set the UI headers
--- a/pkg/ui/handlers.go
+++ b/pkg/ui/handlers.go
@@ -1,0 +1,56 @@
+package ui
+
+import (
+	"fmt"
+	"io/fs"
+	"net/http"
+	"path"
+	"strings"
+
+	"github.com/canonical/identity-platform-admin-ui/internal/logging"
+	"github.com/canonical/identity-platform-admin-ui/internal/monitoring"
+	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
+	"github.com/go-chi/chi/v5"
+)
+
+const UI = "/ui"
+
+type API struct {
+	fileServer http.Handler
+
+	tracer  tracing.TracingInterface
+	monitor monitoring.MonitorInterface
+	logger  logging.LoggerInterface
+}
+
+type Config struct {
+	DistFS fs.FS
+}
+
+func (a *API) RegisterEndpoints(mux *chi.Mux) {
+	mux.Get(fmt.Sprintf("%s/*", UI), a.uiFiles)
+}
+
+// TODO: Validate response when server error handling is implemented
+func (a *API) uiFiles(w http.ResponseWriter, r *http.Request) {
+	// If the path is not /ui{/,} add the html suffix to make file system serving work
+	if ext := path.Ext(r.URL.Path); ext == "" && (r.URL.Path != UI && r.URL.Path != UI+"/") {
+		r.URL.Path = fmt.Sprintf("%s.html", r.URL.Path)
+	}
+
+	r.URL.Path = strings.TrimPrefix(r.URL.Path, UI)
+
+	a.fileServer.ServeHTTP(w, r)
+}
+
+func NewAPI(config *Config, tracer tracing.TracingInterface, monitor monitoring.MonitorInterface, logger logging.LoggerInterface) *API {
+	a := new(API)
+
+	a.fileServer = http.FileServer(http.FS(config.DistFS))
+
+	a.tracer = tracer
+	a.monitor = monitor
+	a.logger = logger
+
+	return a
+}
--- a/pkg/ui/handlers.go
+++ b/pkg/ui/handlers.go
@@ -38,7 +38,20 @@ func (a *API) uiFiles(w http.ResponseWriter, r *http.Request) {
 		r.URL.Path = fmt.Sprintf("%s.html", r.URL.Path)
 	}
 
-	r.URL.Path = strings.TrimPrefix(r.URL.Path, UI)
+	// Set the UI headers
+	// Disables the FLoC (Federated Learning of Cohorts) feature on the browser,
+	// preventing the current page from being included in the user's FLoC calculation.
+	// FLoC is a proposed replacement for third-party cookies to enable interest-based advertising.
+	w.Header().Set("Permissions-Policy", "interest-cohort=()")
+	// Prevents the browser from trying to guess the MIME type, which can have security implications.
+	// This tells the browser to strictly follow the MIME type provided in the Content-Type header.
+	w.Header().Set("X-Content-Type-Options", "nosniff")
+	// Restricts the page from being displayed in a frame, iframe, or object to avoid click jacking attacks,
+	// but allows it if the site is navigating to the same origin.
+	w.Header().Set("X-Frame-Options", "SAMEORIGIN")
+	// Sets the Content Security Policy (CSP) for the page, which helps mitigate XSS attacks and data injection attacks.
+	// The policy allows loading resources (scripts, styles, images, etc.) only from the same origin ('self'), data URLs, and all subdomains of ubuntu.com.
+	w.Header().Set("Content-Security-Policy", "default-src 'self' data: https://*.ubuntu.com; script-src 'self' 'unsafe-inline'; style-src 'self' 'unsafe-inline'")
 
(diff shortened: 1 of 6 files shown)


BarcoMasile (approved):

pkg/ui/handlers.go
```diff
@@ -0,0 +1,56 @@
...
+func NewAPI(config *Config, tracer tracing.TracingInterface, monitor monitoring.MonitorInterface, logger logging.LoggerInterface) *API {
+	a := new(API)
+
+	a.fileServer = http.FileServer(http.FS(config.DistFS))
+
+	a.tracer = tracer
+	a.monitor = monitor
+	a.logger = logger
+
+	return a
```
edlerd, line 55: I think we should add the security headers. See an [example of the headers in LXD-UI](https://github.com/canonical/lxd/blob/371a61a30b8d704aa3982f8bf02054212a209767/lxd/api.go#L86-L95).

pkg/web/router.go
```diff
@@ -148,6 +154,7 @@ func NewRouter(config *RouterConfig, wpool pool.WorkerPoolInterface) http.Handle
 	rulesAPI.RegisterEndpoints(router)
 	rolesAPI.RegisterEndpoints(router)
 	groupsAPI.RegisterEndpoints(router)
+	uiAPI.RegisterEndpoints(router)
```
shipperizer, line 157: this won't fly imho, as all the middlewares for authorisation are hooked already, UI shouldn't be subject to those

edlerd: That's right, the UI should be delivered to any unauthorized requester.

nsklikas: Good point, didn't notice the middleware was there.

Interesting thing I learned about chi router (even though it is not documented):
Using `With` AFTER you register a handler will not affect that handler, so this gives us the behavior we want:
```
	uiAPI.RegisterEndpoints(router)
	router := router.With(...)
        // Register all other handlers
	statusAPI.RegisterEndpoints(apiRouter)
	...
```

Using `With` last causes a panic:
 ```
        // Register all handlers
	uiAPI.RegisterEndpoints(router)
	statusAPI.RegisterEndpoints(apiRouter)
	...
	router := router.With(...)
```

I went with the `Mount` approach and changed the paths on all the handlers, from an API perspective nothing should have changed.

**Comment by nsklikas** [comment]
The frontend is now served on `/`, all the API is served on `/api` (as it used to).

One interesting thing (which I think we should look into) is that the UI changes the path when you click on a component, e.g if you click on provider you will be redirected to `/provider`
image

This means that if I try to refresh the page I will get a `404` because no such page exists.

**Comment by edlerd** [comment]
> One interesting thing (which I think we should look into) is that the UI changes the path when you click on a component, e.g if you click on provider you will be redirected to `/provider`
> 
> This means that if I try to refresh the page I will get a `404` because no such page exists.

In other apps we put the ui paths und /ui and all paths get served as the index file. That file loads some js that handles the current path correctly. Can we do a similar approach here as well?

**Comment by nsklikas** [comment]
> In other apps we serve under /ui and all paths below (like /ui/provider) get served as the same index file. That file loads some js that handles the current path correctly. Can we do a similar approach here as well?

This feels a little wierd to implement in the backend. IMO the backend should not have to implement such logic, but I have no problem with adding it for now.

@BarcoMasile @shipperizer any problem with this?

**Comment by BarcoMasile** [comment]
> > In other apps we serve under /ui and all paths below (like /ui/provider) get served as the same index file. That file loads some js that handles the current path correctly. Can we do a similar approach here as well?
> 
> This feels a little wierd to implement in the backend. IMO the backend should not have to implement such logic, but I have no problem with adding it for now.
> 
> @BarcoMasile @shipperizer any problem with this?

I think @edlerd is referring to the react app handling URL change without actually invoking the backend.
This should be ok doing it in the backend, we need a catch all route to make sure that all "non api" endpoint redirect to the same endpoint that returns the main `index.html` with the compiled react app

**Comment by shipperizer** [comment]
> > In other apps we serve under /ui and all paths below (like /ui/provider) get served as the same index file. That file loads some js that handles the current path correctly. Can we do a similar approach here as well?
> 
> This feels a little wierd to implement in the backend. IMO the backend should not have to implement such logic, but I have no problem with adding it for now.
> 
> @BarcoMasile @shipperizer any problem with this?
> 
> Also @edlerd I am having trouble understanding how that works in the [LXD code](https://github.com/canonical/lxd/blob/371a61a30b8d704aa3982f8bf02054212a209767/lxd/api.go#L86-L95) you sent before, are you sure that they implement this? (e.g. for `/ui/providers` serving `/ui/`)
> 

I'm limited in my knowledge here, but as far as my experience goes with frontends with single page paths were after a `#` in the URL

now I don't have any preference here but one thing I'm not really moving from is that backend logic stays in the backend, frontend logic stays in the frontend
no meddling with boundaries please, unless we have a really good reason about it or this is the best practice and I'm completely off with my reasoning

**Comment by nsklikas** [comment]
> now I don't have any preference here but one thing I'm not really moving from is that backend logic stays in the backend, frontend logic stays in the frontend

That's a good point and that's why I have another issue with this line:
https://github.com/canonical/identity-platform-admin-ui/pull/313/files#diff-f792df7e2bcc8df566b7d3af852ed78b8fd98bef52587eb2ef643ab4429fb7d3R49

It defines a CSP about ubuntu.com and canonical.com, because (I suppose) the UI pulls assets from these servers.
I think that usually in a deployment such headers would be added by the reverse proxy

**Comment by Lukewh** [comment]
My 2c — what @BarcoMasile said. Generally, there will be a catch-all route for all possible UI pages, so when a user visits `/ui/*?` you serve `/ui/index.html` and the SPA will pick up routing for subsequent routing.
We could use the `#/sub-path` approach, but it's largely considered bad practice these days.

You all owe me 2c each.

**Comment by shipperizer** [comment]
> > now I don't have any preference here but one thing I'm not really moving from is that backend logic stays in the backend, frontend logic stays in the frontend
> 
> That's a good point and that's why I have another issue with this line: https://github.com/canonical/identity-platform-admin-ui/pull/313/files#diff-f792df7e2bcc8df566b7d3af852ed78b8fd98bef52587eb2ef643ab4429fb7d3R49
> 
> It defines a CSP about ubuntu.com and canonical.com, because (I suppose) the UI pulls assets from these servers. I think that usually in a deployment such headers would be added by the reverse proxy

the CORS should come via env variable https://github.com/canonical/identity-platform-admin-ui/pull/313/files#diff-e207df27591c5716f54912cf2c0c7e63ffc26755540755cf20ed34fb750f6e04R70 u can use those probably

**Review by edlerd — APPROVED** [review]
Code LGTM.

The Content-Security-Policy might need updating, did you test serving the ui and ensured all pages load as expected and entities can be created / edited / deleted through the forms? I didn't have an easy way to QA.

**Review by shipperizer — CHANGES_REQUESTED** [review]
why did we remove the `api` prefix?

**Comment by shipperizer** [comment]
> why did we remove the `api` prefix?

right I see what happened, let me add some proper comments for a more minimal approach that might work the same

File: pkg/web/router.go [hunk 3]
```diff
@@ -52,6 +55,7 @@ func NewRouter(config *RouterConfig, wpool pool.WorkerPoolInterface) http.Handle
 	idpConfig := config.idp
 	schemasConfig := config.schemas
 	rulesConfig := config.rules
+	uiConfig := config.ui
```
shipperizer, line 58: a couple of lines above we could create the router and then have 
https://pkg.go.dev/github.com/go-chi/chi/v5#Mux.Group to create another inline router for the UI

this would avoid us changing all the `api` prefixes

pkg/web/router.go
```diff
@@ -138,16 +143,20 @@ func NewRouter(config *RouterConfig, wpool pool.WorkerPoolInterface) http.Handle
...
+	clientsAPI.RegisterEndpoints(apiRouter)
+	idpAPI.RegisterEndpoints(apiRouter)
+	schemasAPI.RegisterEndpoints(apiRouter)
+	rulesAPI.RegisterEndpoints(apiRouter)
+	rolesAPI.RegisterEndpoints(apiRouter)
+	groupsAPI.RegisterEndpoints(apiRouter)
+
+	uiAPI.RegisterEndpoints(router)
+
+	router.Mount("/api", apiRouter)
```
shipperizer, line 159: this works and was the part I overlooked in my first comment 

albeit being ok and just right it triggers quite a change in the rest of the code which I would probably spare

on a side note, we would also need to have a catch all to point to the UI handler and we could use https://pkg.go.dev/github.com/go-chi/chi/v5#Mux.NotFound
on top of still having the UI served from `/ui`

nsklikas:
> The Content-Security-Policy might need updating, did you test serving the ui and ensured all pages load as expected and entities can be created / edited / deleted through the forms? I didn't have an easy way to QA.

This is exactly why it's not a good idea to have the backend setting these headers, I tested all the pages and it seems to work fine.

pkg/web/router.go
```diff
@@ -124,9 +127,14 @@ func NewRouter(config *RouterConfig, wpool pool.WorkerPoolInterface) http.Handle
 		logger,
 	)
 
+	// apply authorization and validation middlewares using With to overcome issue with <id> URLParams not available
+	apiRouter := router.Group(func(r chi.Router) {
+		r.Use(authorizationMiddleware)
```
shipperizer, line 132: let's validate this works correctly due to how route params get injected (that's why we were using `With`)

edlerd:
> This is exactly why it's not a good idea to have the backend setting these headers, I tested all the pages and it seems to work fine.

Thanks for checking!

I don't have a better approach, as the UI code can't control or set the headers. I think the UI needs to depend on the backend for the security headers, we can't remove the dependency here.

pkg/web/router.go
```diff
@@ -124,9 +127,14 @@ func NewRouter(config *RouterConfig, wpool pool.WorkerPoolInterface) http.Handle
...
+	}).(*chi.Mux)
+
 	if config.payloadValidationEnabled {
 		validationRegistry := validation.NewRegistry(tracer, monitor, logger)
-		router = router.With(validationRegistry.ValidationMiddleware).(*chi.Mux)
+		apiRouter = apiRouter.With(validationRegistry.ValidationMiddleware).(*chi.Mux)
```
nsklikas, line 137: Should I change this to use `Use` as well to keep it consistent?

nsklikas: I could also switch the other one back to `With`

shipperizer (dismissed):

nsklikas:
- Switched other `With` to `Use`
- Change the ui handler to not return the index page in case of unknown routes that start with `/api`

Should be good to go

shipperizer (dismissed):

pkg/ui/handlers.go
```diff
@@ -0,0 +1,70 @@
+package ui
```
shipperizer, line 1: add copyrights

nsklikas:
After testing it with traefik, I moved all the ui assets back to `/ui`, the reason is we cannot tell if the URL that the user used has a trailing slash or not for the root url (due to traefik re-writing the path).

To serve the UI under the root path, traefik would have to provide a middleware that adds a trailing slash to all URLs if it is missing (it is possible, but the charm does not implement it).

The index page now works fine, but the UI needs to fix its pathing so that it works with relative paths. This may cause new issue to come to light.

BarcoMasile (dismissed):

shipperizer (approved):

PR #314, opened 2024-06-06
IAM 885 OAuth2 client and resource server integration

BarcoMasile:


nsklikas (commented):
Left some comments, I realize that this is a temporary solution and adding cookie support will change the implementation so some comments may not make much sense. 

Feel free to disregard most of my comments, but I think that the handler should  be broken in 2.

pkg/authentication/context.go
```diff
@@ -0,0 +1,127 @@
...
+
+	tracer trace.Tracer
+	logger logging.LoggerInterface
+}
+
+func (o *OAuth2Context) LoginRedirect(w http.ResponseWriter, r *http.Request) {
+	_, span := o.tracer.Start(r.Context(), "authentication.OAuth2Context.LoginRedirect")
+	defer span.End()
+
+	additionalInfo, _ := json.Marshal(map[string]string{"referer": r.Referer(), "remote-address": r.RemoteAddr})
```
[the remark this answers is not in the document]
BarcoMasile: This is because Hydra rejects empty state parameter and needs a certain amount of characters to have enough entropy. I used this two values because it should be more performant than generating pseudorandomic values and because we could actually use those for auditing/logging

pkg/authentication/middlewares.go
```diff
@@ -0,0 +1,93 @@
...
+	"net/http"
+	"strings"
+
+	"github.com/canonical/identity-platform-admin-ui/internal/http/types"
+	"github.com/canonical/identity-platform-admin-ui/internal/logging"
+	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
+)
+
+type Middleware struct {
+	allowListedEndpoints []string
```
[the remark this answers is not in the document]
BarcoMasile: Great suggestion, thanks!

pkg/authentication/handlers.go
```diff
@@ -0,0 +1,109 @@
...
+
+	tracer trace.Tracer
+	logger logging.LoggerInterface
+}
+
+func (a *API) RegisterEndpoints(mux *chi.Mux) {
+	mux.Get("/api/v0/login", a.handleLogin)
+}
+
+func (a *API) handleLogin(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
BarcoMasile: I like the suggestion, I would prefer doing it in another PR (like when we switch to cookies, since I'll have to change the tests and implementation anyway, and I could spare it now)

pkg/authentication/tokens.go
```diff
@@ -0,0 +1,122 @@
...
+
+	"github.com/coreos/go-oidc/v3/oidc"
+	"go.opentelemetry.io/otel/trace"
+	"golang.org/x/oauth2"
+
+	"github.com/canonical/identity-platform-admin-ui/internal/logging"
+)
+
+type JWKSTokenVerifier struct {
+	clientID string
```
wood-push-melon, line 18: Just a **NONBLOCKING** question (probably I miss some knowledge context): why does the verifier have an attribute `clientID`? Did not see where it is used for.

pkg/authentication/context.go
```diff
@@ -0,0 +1,127 @@
...
+func (o *OAuth2Context) Verifier() TokenVerifier {
+	return o.verifier
+}
+
+func NewOAuth2Context(config *Config, getProvider OIDCProviderSupplier, tracer trace.Tracer, logger logging.LoggerInterface) *OAuth2Context {
+	o := new(OAuth2Context)
+	o.tracer = tracer
+	o.logger = logger
+
+	ctx := oidc.InsecureIssuerURLContext(context.Background(), config.issuer)
```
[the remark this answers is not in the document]
BarcoMasile: Confirmed, you're right. It was actually the leftover. I removed it. Thanks!

pkg/authentication/tokens.go
BarcoMasile: You are right, I put it there for consistency with the other verifier, and with the thought in mind that it could be helpful for next iterations. But I'll remove it, and if needed we'll add it back. Good catch!

BarcoMasile: I just pushed the change.

pkg/authentication/context.go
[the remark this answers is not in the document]
nsklikas: Oh this is used for the state, I didn't notice. The state needs to be random, I don't think that this is enough.

I had a look at the library we use and it turns out that it does not validate the value of the state on the callback. We need to add that (and we need to start storing the state in the cookie).

To add to this we should also send a random `nonce` to the provider and the ID Token that we get back must have the same nonce parameter.

nsklikas (changes requested):
Perhaps we should consider using https://github.com/coreos/go-oidc, it is a wrapper around the oauth2 std lib (the one that we use). Jimm went with this one and I think it would be nice if could have a consistent stack.

TBH I don't love either of those implementations

pkg/authentication/tokens.go
```diff
@@ -0,0 +1,120 @@
...
+
+type claims struct {
+	Audience []string `json:"aud"`
+}
+
+func (u *UserinfoTokenVerifier) VerifyAccessToken(ctx context.Context, rawAccessToken string) (*Principal, error) {
+	_, span := u.tracer.Start(ctx, "authentication.UserinfoTokenVerifier.VerifyAccessToken")
+	defer span.End()
+
+	info, err := u.provider.UserInfo(ctx, oauth2.StaticTokenSource(&oauth2.Token{AccessToken: rawAccessToken}))
```
shipperizer, line 73: same here for otelhttp, we will need to add something to the context to make sure it s using a traceable http client

shipperizer (commented):
there are quite a lot of layers of abstractions which took me a while to understand but all in all it seems to be doing what it's supposed to

pkg/authentication/handlers.go
```diff
@@ -0,0 +1,109 @@
...
+	scopes               []string `validate:"required,dive,required"`
+}
+
+type oauth2Tokens struct {
+	IDToken      string `json:"id_token"`
+	AccessToken  string `json:"access_token"`
+	RefreshToken string `json:"refresh_token"`
+}
+
+func NewAuthenticationConfig(specs *config.EnvSpec) *Config {
```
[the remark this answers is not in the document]
BarcoMasile: I understand the tradeoff, I was trying to simplify the value carrying but I'm open. I'll change it

pkg/authentication/interfaces.go
```diff
@@ -0,0 +1,40 @@
...
+	"github.com/coreos/go-oidc/v3/oidc"
+	"golang.org/x/oauth2"
+)
+
+type providerVerifier interface {
+	Verify(ctx context.Context, rawIDToken string) (*oidc.IDToken, error)
+}
+
+type TokenVerifier interface {
+	VerifyAccessToken(ctx context.Context, rawAccessToken string) (*Principal, error)
```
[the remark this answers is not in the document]
BarcoMasile: I'm open to drop it since it's clear anyway
I would usually prefer to have names when multiple parameters are present like multiple strings, to allow distinguishing.
Changing it in next revision

pkg/authentication/context.go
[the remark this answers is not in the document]
BarcoMasile: As much as I agree with the use of the nonce parameter to avoid replay attacks, I consider that a nice improvement, but an improvement nonetheless, so I am glad to work on that, just not in this PR since it's definitely out scope.
About the state, hydra makes it mandatory, it usually isn't. I chose those values to satisfy hydra's requirements.
I would actually use the state parameter to provide a CSRF protection layer, which also would be too much for this first integration.

I am open to change this state value to something pseudo randomic though, and I'm about to do it in next push.
Thanks!

pkg/authentication/tokens.go
```diff
@@ -0,0 +1,120 @@
...
+	"fmt"
+
+	"github.com/coreos/go-oidc/v3/oidc"
+	"go.opentelemetry.io/otel/trace"
+	"golang.org/x/oauth2"
+
+	"github.com/canonical/identity-platform-admin-ui/internal/logging"
+)
+
+type JWKSTokenVerifier struct {
```
shipperizer, line 17: let's add the monitor here as we might want to add metrics re validation

nsklikas:
> We are using exactly that library (see [go.mod](https://github.com/canonical/identity-platform-admin-ui/blob/270b05951a7d77645fe1c2ea386cd214e0da932c/go.mod)), which provides only the OIDC layer on top of the OAuth2 "layer" provided by a 3rd party library.

You are right, I am blind. I was confused because a lot of the methods/objects come from the std lib

pkg/authentication/handlers.go
```diff
@@ -0,0 +1,109 @@
...
+		_ = json.NewEncoder(w).Encode(
+			types.Response{
+				Status:  http.StatusBadRequest,
+				Message: err.Error(),
+			},
+		)
+		return
+	}
+
+	rawIDToken := oauth2Token.Extra("id_token").(string)
```
nsklikas, line 87: We should verify the ID token (similar to https://github.com/coreos/go-oidc/blob/v3/example/idtoken/app.go#L103)

BarcoMasile: I'll do it the next push (coming up), thanks

pkg/authentication/context.go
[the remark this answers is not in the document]
BarcoMasile: I changed my mind and implemented the nonce check in this PR, so I'll push the changes as soon as I'm done addressing all other comments. Thanks!

pkg/authentication/middlewares.go
```diff
@@ -0,0 +1,93 @@
...
+		if m.isAllowListed(r) {
+			next.ServeHTTP(w, r.WithContext(ctx))
+			return
+		}
+
+		rawAccessToken, err := getBearerToken(r.Header)
+		if err != nil {
+			m.unauthorizedResponse(w, err)
+			return
+		}
```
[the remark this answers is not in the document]
BarcoMasile: If that's the plan then we should talk mroe about this, since with the cookie based authentication we could rely on multiple pieces of info (id, access and refresh tokens), so we could (for users) refresh the access token if it's expired.
Let's talk about this and provide a plan for the future

pkg/authentication/context.go
```diff
@@ -0,0 +1,125 @@
...
+	o := new(OAuth2Context)
+	o.tracer = tracer
+	o.logger = logger
+
+	provider, err := getProvider(context.Background(), config.issuer)
+	if err != nil {
+		o.logger.Fatalf("Unable to fetch provider info, error: %v", err.Error())
+	}
+
+	var verifier TokenVerifier
```
wood-push-melon, line 104: Just a **NONBLOCKING** question: do you think it's better to move this token verifier "factory" logic to `tokens.go`? It seems that `OAuth2Context` does not need to care about how to create an appropriate token verifier. It only needs to use the factory to get the right one.

internal/config/specs.go
```diff
@@ -23,6 +23,14 @@ type EnvSpec struct {
 	HydraAdminURL       string `envconfig:"hydra_admin_url" required:"true"`
 	OathkeeperPublicURL string `envconfig:"oathkeeper_public_url" required:"true"`
 
+	AuthenticationEnabled           bool     `envconfig:"authentication_enabled" default:"false" validate:"required"`
+	OIDCIssuer                      string   `envconfig:"oidc_issuer" validate:"required"`
+	OAuth2ClientId                  string   `envconfig:"oauth2_client_id" validate:"required"`
+	OAuth2ClientSecret              string   `envconfig:"oauth2_client_secret" validate:"required"`
+	OAuth2RedirectURI               string   `envconfig:"oauth2_redirect_uri" validate:"required"`
+	OAuth2CodeGrantScopes           []string `envconfig:"oauth2_codegrant_scopes" default:"openid,offline_access" validate:"required"`
+	AccessTokenVerificationStrategy string   `envconfig:"access_token_verification_strategy" default:"jwks" validate:"oneof=jwks userinfo"`
```
nsklikas, line 32: Should we have a common prefix for the oauth2 vars?

pkg/authentication/context.go
BarcoMasile: You are right and that was my original implementation, but since the factory process is so simple, I removed the factory classt and just made it part of the OAuth2Context, so it's one less object to crowd the implementation.
We can still extract it in case things get more complicated.

internal/config/specs.go
BarcoMasile: I tried to keep it consistent, thing is, the only one missing the oauth2 prefix is the verification strategy (which already has a long name) so I left it like that since "access token" is already explicitly oauth2 related.
OIDC issuer I prefer to keep it OIDC prefixed for its meaning, and the first AuthenticationEnabled I didn't put the prefix there because authn is at a higher conceptual level that OAuth2.
So basically I could add the prefix to the verification strategy making it `OAuth2AccessTokenVerificationStrategy`, I'm open to it

pkg/authentication/context.go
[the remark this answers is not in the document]
BarcoMasile: In the PR change coming up today you'll see the nonce implementation. Thing is, nonce should be encrypted, but encryption will come with next story. So for now it will be base64 encoded, and a todo is there to track that (it will be addressed very soon so no need for a new issue).

pkg/authentication/cookies.go
```diff
@@ -0,0 +1,40 @@
...
+	NONCE_COOKIE_NAME = "nonce"
+)
+
+func SetNonceCookie(w http.ResponseWriter, nonce string) {
+	expires := time.Now().Add(2 * time.Minute)
+	http.SetCookie(w, &http.Cookie{
+		Name: NONCE_COOKIE_NAME,
+		// TODO @barco: encrypt this value when cookie encryption is added to the codebase
+		Value:    nonce,
+		Path:     "/api/v0/login",
```
nsklikas, line 22: I may be wrong, but I am not sure if this will work when running under a path, e.g. the login URL may be:
`https://example.com/iam-admin/api/v0/login`, afaict in that case this cookie path won't work

pkg/authentication/context.go
```diff
@@ -0,0 +1,156 @@
...
+	monitor monitoring.MonitorInterface
+	rand    *rand.Rand
+}
+
+func (o *OAuth2Context) LoginRedirect(w http.ResponseWriter, r *http.Request) {
+	_, span := o.tracer.Start(r.Context(), "authentication.OAuth2Context.LoginRedirect")
+	defer span.End()
+
+	nonce, _ := nonce(r)
+	state := o.state(stateSize)
```
nsklikas, line 84: Since cookie support was added for `nonce`, shouldn't we add the same verification logic for `state`?

pkg/authentication/context.go
```diff
@@ -0,0 +1,156 @@
...
+
+	SetNonceCookie(w, nonce)
+
+	http.Redirect(
+		w, r,
+		o.client.AuthCodeURL(state, oidc.Nonce(nonce), oauth2.SetAuthURLParam("audience", o.client.ClientID)),
```
nsklikas, line 90: Can we add a TODO to remove this once https://github.com/canonical/identity-platform-login-ui/issues/244 is fixed?

pkg/authentication/context.go
```diff
@@ -0,0 +1,156 @@
...
+func (o *OAuth2Context) state(size int) string {
+	state := make([]rune, size)
+	for i := range state {
+		state[i] = letterRunes[rand.Intn(len(letterRunes))]
+	}
+	return string(state)
+}
+
+func nonce(r *http.Request) (string, error) {
+	marshal, err := json.Marshal(map[string]string{"referer": r.Referer(), "remote-address": r.RemoteAddr})
```
nsklikas, line 104: Since we have a random generator for state, why don't we reuse that for nonce as well? Unless we have a use for these values.

pkg/authentication/cookies.go
```diff
@@ -0,0 +1,40 @@
...
+	"time"
+)
+
+var (
+	EPOCH             = time.Unix(0, 0)
+	NONCE_COOKIE_NAME = "nonce"
+)
+
+func SetNonceCookie(w http.ResponseWriter, nonce string) {
+	expires := time.Now().Add(2 * time.Minute)
```
nsklikas, line 17: I feel like this should be larger, maybe we should make it configurable?

pkg/authentication/cookies.go
BarcoMasile: You're not wrong. Right now every handler endpoint is hardcoded and does not take into account any prefix. When we implement that, we will have to take care of that in all the codebase.

pkg/authentication/tokens.go
BarcoMasile: Done in both token verifiers

pkg/authentication/tokens.go
BarcoMasile: Done for every `context` use in the OAuth2 implementation. The httpclient with Otel transport is shared.

nsklikas (commented):
I understand that this is not a complete set of changes and some issues with this codebase will be addressed on later PRs. I think it would make approving this easier if we could have a list of things that need to be done once this PR is merged

pkg/authentication/cookies.go
BarcoMasile: Making it configurable right now

pkg/authentication/cookies.go
shipperizer: ingress controller will take care of doing the path rewriting, now do we actually care in this case what s inside the cookie given that the path of the request will be rewritten to what we know?

pkg/authentication/context.go
BarcoMasile: I would honestly skip this for now, since I already adeded that check it should be more than enough for a first implementation. We can add/change that whenever we want. My idea would be to make it part of a CSRF protection implementation we can add in the future.

pkg/authentication/cookies.go
BarcoMasile: We could "ignore" the path although it's nice to keep this cookie "scoped" to the only endpoint that needs it.
My proposal is: leave it as is. With cookie encryption implemented, we would have our secret key as an additional strong layer of security and could decide to relax the `path` constraint if it troubles us with the deployment

pkg/authentication/context.go
BarcoMasile: I would skip this additional change since it would require me to bring a bunch of changes to make it testable, and would cause an avalanche effect from addressing comments

pkg/authentication/context.go
nsklikas: I have no problem not to add this in this PR to unblock development, but we *need* to add state verification eventually. 

Several attacks described in https://datatracker.ietf.org/doc/html/rfc6819 can be mitigated with the `state` parameter.

pkg/authentication/context.go
nsklikas: Same here. I have no problem with keeping this as it is for now, as long as we have a clear backlog of what needs to be done and that we eventually address this.

We should use random values for `nonce` and `state`. Using headers like `Referer` and `RemoteAddr` opens the window for attacks, because it is very easy for someone to tamper with them

pkg/authentication/context.go
```diff
@@ -0,0 +1,168 @@
...
+const stateSize = 18
+
+var (
+	letterRunes         = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
+	PrincipalContextKey principalContextKey
+	otelHTTPClient      = http.Client{Transport: otelhttp.NewTransport(http.DefaultTransport)}
+)
+
+type Principal struct {
+	Subject string `json:"sub"`
```
nsklikas, line 34: Question out of scope for this PR: Shouldn't we use the user's email instead of the `sub`? This would be needed for the exposing the user's ID in the UI

pkg/authentication/context.go
BarcoMasile: [Issue](https://github.com/canonical/identity-platform-admin-ui/issues/321) for this

pkg/authentication/context.go
BarcoMasile: Totally agree on adding state check in the near future.
[Issue](https://github.com/canonical/identity-platform-admin-ui/issues/322) to track this.

pkg/web/router.go
```diff
@@ -132,6 +137,16 @@ func NewRouter(config *RouterConfig, wpool pool.WorkerPoolInterface) http.Handle
...
 	}).(*chi.Mux)
 
+	var oauth2Context *authentication.OAuth2Context
+
+	if oauth2Config.Enabled {
+		oauth2Context = authentication.NewOAuth2Context(config.oauth2, oidc.NewProvider, tracer, logger, monitor)
+
+		authenticationMiddleware := authentication.NewAuthenticationMiddleware(oauth2Context, tracer, logger)
+		authenticationMiddleware.SetAllowListedEndpoints("/api/v0/login", "/api/v0/status", "api/v0/metrics")
+		apiRouter = apiRouter.With(authenticationMiddleware.OAuth2Authentication).(*chi.Mux)
```
nsklikas, line 147: I switched the other routers to use `Use`, should we change this one as well? (the issue that we had with `Use` does not happen with subrouters)

nsklikas (dismissed):
minor comment, but approved

nsklikas (approved):

PR #324, opened 2024-06-13
IAM 921 add state check, improve nonce generation, separate login handlers and improve tests

### LINKED ISSUE #321: Switch a pseudorandomic implementation for the `nonce` value generation + make it testable
## Description
Right now the nonce value is generated using current request referrer and IP address. We should switch to a pseudorandomicly generated value like the state.

A new interface would be needed to make it testable.

### LINKED ISSUE #322: Add `state` verification in login process
## Description
Right now only `nonce` value gets checked during the login process. We need to also add `state` parameter verification. Hopefully as part of a CSRF protection implementation.

BarcoMasile:
## Description
This PR introduces the state check, and makes nonce and state values generation UUID based. This is due to a benchmark that compared UUIDs (v4) generation to random string (of the same lenght) generation from the previous implementation.
Results are not incredibly far from each other but still UUID proved to be slightly better performant.

This PR also changes the redirect uri from `/api/v0/login` to `/api/v0/auth/callback`, because the original login handler is now split into two different handlers.
The login init endpoint is now `/api/v0/auth`.

### Benchmark results:
```
BenchmarkOAuth2Helper_RandomURLString
BenchmarkOAuth2Helper_RandomURLString-12    	 2434296	       450.7 ns/op
BenchmarkOAuth2Helper_RandomURLString-12    	 2259633	       464.3 ns/op
BenchmarkOAuth2Helper_RandomURLString-12    	 2562487	       461.3 ns/op
BenchmarkOAuth2Helper_RandomURLString-12    	 2551335	       480.2 ns/op
BenchmarkOAuth2Helper_RandomURLString-12    	 2511240	       501.0 ns/op
BenchmarkOAuth2Helper_UUID
BenchmarkOAuth2Helper_UUID-12               	 4075657	       310.3 ns/op
BenchmarkOAuth2Helper_UUID-12               	 4259341	       295.9 ns/op
BenchmarkOAuth2Helper_UUID-12               	 3992308	       299.4 ns/op
BenchmarkOAuth2Helper_UUID-12               	 4207932	       295.7 ns/op
BenchmarkOAuth2Helper_UUID-12               	 3598333	       313.4 ns/op
```

## Changes
- **NB: redirect URI is now `/api/v0/auth/callback`**
- switched from `OAuth2NonceCookieTTL` to `OAuth2AuthCookiesTTL` (env var is now `OAUTH2_AUTH_COOKIES_TTL`)
- externalized nonce and state generation in helper object
- added auth cookie manager for nonce and state cookies handling
- login and token retrieval handler is now separated in two different handlers
- improved the handlers error overall
- improved logging

### Missing
- helper tests
- auth cookie manager tests

### Issues
Closes #321 
Closes #322


internal/config/specs.go
```diff
@@ -32,7 +32,7 @@ type EnvSpec struct {
 	OAuth2ClientSecret              string        `envconfig:"oauth2_client_secret" validate:"required"`
 	OAuth2RedirectURI               string        `envconfig:"oauth2_redirect_uri" validate:"required"`
 	OAuth2CodeGrantScopes           []string      `envconfig:"oauth2_codegrant_scopes" default:"openid,offline_access" validate:"required"`
-	OAuth2NonceCookieTTL            time.Duration `envconfig:"oauth2_nonce_cookie_ttl" default:"5m0s" validate:"required"`
+	OAuth2AuthCookiesTTL            time.Duration `envconfig:"oauth2_auth_cookies_ttl" default:"5m0s" validate:"required"`
```
shipperizer, line 35: as env var this is a bit non standard in terms of value, i would say it's easier to just have it in seconds so it becomes a plain number (and hopefully we would be able to import it and parse)

i cant really see `OAUTH2_AUTH_COOKIES_TTL=30m` being easy to use, but happy to stick with it if parsing is to hard to implement right now

pkg/authentication/handlers.go
```diff
@@ -17,15 +17,38 @@ import (
...
+)
+
+var (
+	codeParamNotFoundErr   = errors.New(codeParamNotFoundMsg)
+	stateParamNotFoundErr  = errors.New(stateParamNotFoundMsg)
+	noIDTokenErr           = errors.New(noIDTokenMsg)
+	nonceCookieNotFoundErr = errors.New(nonceCookieNotFoundMsg)
+	idTokenNonceMatchErr   = errors.New(idTokenNonceMatchMsg)
+	stateCookieNotFoundErr = errors.New(stateCookieNotFoundMsg)
+	stateParamMatchErr     = errors.New(stateParamMatchMsg)
```
shipperizer, line 40: why `errors.New` and not `fmt.Errorf`?second one is used more across the codebase

i've got a tiny issue with global vars, as they can be changed randomly in the codebase and we would get unpredictable results

can we instantiate new objects and return them instead?

pkg/authentication/handlers.go
```diff
@@ -53,26 +76,57 @@ type API struct {
...
 	oauth2           OAuth2ContextInterface
+	helper           OAuth2HelperInterface
+	authCookiesTTL   time.Duration
 
-	tracer   trace.Tracer
-	logger   logging.LoggerInterface
-	nonceTTL time.Duration
+	tracer        trace.Tracer
+	logger        logging.LoggerInterface
+	cookieManager AuthCookieManagerInterface
```
shipperizer, line 84:
```suggestion
	cookieManager AuthCookieManagerInterface
	
	tracer        trace.Tracer
	logger        logging.LoggerInterface
```

just a matter of standardization, o11y stuff is always in the bottom

pkg/authentication/handlers.go
```diff
@@ -53,26 +76,57 @@ type API struct {
...
 	r = r.WithContext(OtelHTTPClientContext(r.Context()))
 
-	code := r.URL.Query().Get("code")
+	nonce := a.helper.RandomURLString()
+	state := a.helper.RandomURLString()
+
+	a.cookieManager.SetNonceCookie(w, nonce, a.authCookiesTTL)
+	a.cookieManager.SetStateCookie(w, state, a.authCookiesTTL)
+
+	a.oauth2.LoginRedirect(w, r, nonce, state)
```
shipperizer, line 102: not sure u need to pass them (`nonce` and `state`) in as u just wrote them in the response, can we grab them from there?

pkg/authentication/handlers.go
```diff
@@ -86,8 +140,8 @@ func (a *API) handleLogin(w http.ResponseWriter, r *http.Request) {
 
 	rawIDToken, ok := oauth2Token.Extra("id_token").(string)
 	if !ok {
-		a.logger.Error("unable to retrieve ID token")
-		badRequest(w, errors.New("unable to retrieve ID token"))
+		a.logger.Error(noIDTokenMsg)
+		badRequest(w, noIDTokenErr)
```
shipperizer, line 144: missed this in the prev pr, but basically (and this is a matter of preference) handler code is quite limited and i do believe there is some benefit in having it as explicit as possible

now having a helper function is great but at the same time makes u have another jump in the codebase to understand what's going on

i'm happy to leave it there but unless it get's used across the board in all the API i would say that in this case just makes it a bit more hideous to understand what's going on compared to any other pkg handlers

pkg/authentication/handlers.go
```diff
@@ -123,6 +169,37 @@ func (a *API) handleLogin(w http.ResponseWriter, r *http.Request) {
...
+		a.logger.Error(stateCookieNotFoundMsg)
+		return stateCookieNotFoundErr
+	}
+
+	if stateCookieValue != state {
+		a.logger.Error(stateParamMatchMsg)
+		return stateParamMatchErr
+	}
+
+	return nil
```
shipperizer, line 202: i wonder if this should be deemed "business logic" and moved to a `service.go`

pkg/authentication/handlers.go
```diff
@@ -136,12 +213,15 @@ func badRequest(w http.ResponseWriter, err error) {
 	return
 }
 
-func NewAPI(oauth2Context OAuth2ContextInterface, tracer trace.Tracer, logger logging.LoggerInterface) *API {
+func NewAPI(oauth2Context OAuth2ContextInterface, helper OAuth2HelperInterface, cookieManager AuthCookieManagerInterface, authCookiesTTL time.Duration, tracer trace.Tracer, logger logging.LoggerInterface) *API {
```
shipperizer, line 216:
```suggestion
func NewAPI(authCookiesTTL time.Duration, oauth2Context OAuth2ContextInterface, helper OAuth2HelperInterface, cookieManager AuthCookieManagerInterface, tracer trace.Tracer, logger logging.LoggerInterface) *API {
```

usually we put simpler objects first

pkg/authentication/handlers.go
```diff
@@ -123,6 +169,37 @@ func (a *API) handleLogin(w http.ResponseWriter, r *http.Request) {
...
+
+	if idToken.Nonce != nonce {
+		a.logger.Error(idTokenNonceMatchMsg)
+		return idTokenNonceMatchErr
+	}
+
+	return nil
+}
+
+func (a *API) stateCheck(r *http.Request, state string) error {
```
nsklikas, line 190: nitpick: I prefer `checkState`, that way is it is clearer that this function checks something

pkg/authentication/handlers.go
```diff
@@ -123,6 +169,37 @@ func (a *API) handleLogin(w http.ResponseWriter, r *http.Request) {
 	}
 
 	_ = json.NewEncoder(w).Encode(tokens)
+
+}
+
+func (a *API) nonceCheck(r *http.Request, idToken *Principal) error {
```
nsklikas, line 175: nitpick: I prefer `checkNonce`, that way is it is clearer that this function checks something

pkg/web/router.go
```diff
@@ -143,7 +143,12 @@ func NewRouter(config *RouterConfig, wpool pool.WorkerPoolInterface) http.Handle
 		oauth2Context = authentication.NewOAuth2Context(config.oauth2, oidc.NewProvider, tracer, logger, monitor)
 
 		authenticationMiddleware := authentication.NewAuthenticationMiddleware(oauth2Context, tracer, logger)
-		authenticationMiddleware.SetAllowListedEndpoints("/api/v0/login", "/api/v0/status", "api/v0/metrics")
+		authenticationMiddleware.SetAllowListedEndpoints(
+			"/api/v0/login",
+			"/api/v0/auth/callback",
+			"/api/v0/status",
+			"api/v0/metrics",
```
nsklikas, line 150:
```suggestion
			"/api/v0/metrics",
```

pkg/authentication/handlers.go
```diff
@@ -53,26 +76,57 @@ type API struct {
...
 func (a *API) handleLogin(w http.ResponseWriter, r *http.Request) {
 	// add the Otel HTTP Client
 	r = r.WithContext(OtelHTTPClientContext(r.Context()))
 
-	code := r.URL.Query().Get("code")
+	nonce := a.helper.RandomURLString()
+	state := a.helper.RandomURLString()
+
+	a.cookieManager.SetNonceCookie(w, nonce, a.authCookiesTTL)
+	a.cookieManager.SetStateCookie(w, state, a.authCookiesTTL)
```
nsklikas, line 100: question: does this belong on the handler layer? I think that authn request param generation should be done in another place

pkg/authentication/handlers.go
```diff
@@ -17,15 +17,38 @@ import (
...
+	codeParamNotFoundMsg   = "OAuth2 code not found"
+	stateParamNotFoundMsg  = "OAuth2 state not found"
+	noIDTokenMsg           = "unable to retrieve ID token"
+	nonceCookieNotFoundMsg = "nonce cookie not found"
+	idTokenNonceMatchMsg   = "id token nonce does not match nonce cookie"
+	stateCookieNotFoundMsg = "state cookie not found"
+	stateParamMatchMsg     = "state parameter does not match state cookie"
+)
+
+var (
```
nsklikas, line 33: Use `const` instead?

nsklikas (commented):
Mostly minor comments, one thing that I am not sure how I feel about is the code structure (handlers feel overloaded, not sure if helpers make sense and not sure if `context.go` is a good name for a file that contains our oidc implementation)

internal/config/specs.go
BarcoMasile: Duration parsing is usually pretty standard for extrenalized configs, but I can change it OAuth2AuthCookieTTLSeconds and make it an int, no problem.
I'll push this change with the missing texts as soon as I have them

pkg/authentication/handlers.go
BarcoMasile: I believe this function doesn't have to "know" that those values are in there alreadyI wouldn't like that, also I prefer to keep it direct by passing them explicitly

pkg/authentication/handlers.go
BarcoMasile: I don't agree because this being "request validation" it belongs to the controller layer.
The only "out of tone" parte is checking the IDToken "nonce" claim but it wouldn't be worth it to separate that if check

pkg/web/router.go
BarcoMasile: Thanks

pkg/authentication/handlers.go
BarcoMasile: The generation is in fact delegated to the helper object.
Setting cookies is 100% controller layer logic so it belongs in the handler, that's why the controller uses the helper.

pkg/authentication/handlers.go
BarcoMasile: I wish!
Unfortunately the "amazing" Go language doesn't allow to have constant errors.

BarcoMasile:
> Mostly minor comments, one thing that I am not sure how I feel about is the code structure (handlers feel overloaded, not sure if helpers make sense and not sure if `context.go` is a good name for a file that contains our oidc implementation)

I'm open to suggestions, probably `oauth2.go` or `oidc.go` could be good. I'll change the name change along with the addressing all the other comments

pkg/authentication/handlers.go
nsklikas: TIL

pkg/authentication/handlers.go
```diff
@@ -44,35 +49,68 @@ func NewAuthenticationConfig(enabled bool, issuer, clientID, clientSecret, redir
...
-	code := r.URL.Query().Get("code")
+	nonce := a.helper.RandomURLString()
+	state := a.helper.RandomURLString()
+
+	ttl := time.Duration(a.authCookiesTTLSeconds) * time.Second
+
+	a.cookieManager.SetNonceCookie(w, nonce, ttl)
+	a.cookieManager.SetStateCookie(w, state, ttl)
+
+	a.oauth2.LoginRedirect(w, r, nonce, state)
```
nsklikas, line 86: nitpick: I would prefer if this function only returned the URL and the handler wrote to the response the redirect. Mainly for clear separation of responsibility. No problem with keeping it as it is.

pkg/authentication/handlers.go
```diff
@@ -44,35 +49,68 @@ func NewAuthenticationConfig(enabled bool, issuer, clientID, clientSecret, redir
...
+	cookieManager         AuthCookieManagerInterface
+	authCookiesTTLSeconds int
+
+	tracer trace.Tracer
+	logger logging.LoggerInterface
 }
 
 func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	mux.Get("/api/v0/login", a.handleLogin)
+	mux.Get("/api/v0/auth/callback", a.handleCallback)
```
nsklikas, line 71: thought: would we want to have a common prefix for these? (e.g. `/api/v0/auth` and `/api/v0/auth/callback`)

nsklikas (dismissed):
a couple of nitpicks for me, feel free to disregard if they don't make much sense

pkg/authentication/handlers.go
BarcoMasile: Yes it makes sense, I actually thought about it but forgot to act on it. I was thinking /auth/login and /auth/callback
but also your suggestion is fine for me. I'll change it

shipperizer (approved):

nsklikas (approved):

PR #329, opened 2024-06-17
feat: add encryption mechanism for auth cookies

BarcoMasile:
## Description
This PR introduces both the interface `EncryptInterface` and its implementation. This is needed to allow symmetric encryption of string values.
This PR also adopts this newly added object in the `AuthCookieManager` without changing its methods signatures, to get us one step closer to the implementation of the user session spec, which will happen during the rest of the activities in [IAM-811](https://warthogs.atlassian.net/browse/IAM-811).

The new secret key for cookie values encryption must be exactly 32 bytes, as specified in the [OAUTH2.md](https://github.com/canonical/identity-platform-admin-ui/blob/7b64ac5b43db6186e2b502e6a199553a4f38394e/OAUTH2.md) doc.

## Important note on symmetric encryption
The choice in this PR was to use the standard library, as the preferred strategy throughout the codebase. The standard library implementation of the AES algorithm with nonce allows for 16, 24 or 32 bytes for the secret key. The choice here is to go with 32 bytes.

[IAM-811]: https://warthogs.atlassian.net/browse/IAM-811?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/encryption/encryption.go
```diff
@@ -0,0 +1,79 @@
...
+	logger logging.LoggerInterface
+}
+
+// Encrypt takes a plain string and returns a hex encoded string
+func (e *Encrypt) Encrypt(data string) (string, error) {
+	payload := []byte(data)
+	nonce, err := e.generateCipherNonce()
+	if err != nil {
+		e.logger.Errorf("error generating random the nonce, %v", err)
+		return "", fmt.Errorf("error generating random the nonce, %v", err)
```
shipperizer, line 29:
```suggestion
     err = fmt.Errorf("error generating random the nonce, %v", err)
		e.logger.Errorf(err.Error())
		return "", err
```

this might do too

internal/encryption/encryption.go
```diff
@@ -0,0 +1,79 @@
...
+	"fmt"
+	"io"
+
+	"github.com/canonical/identity-platform-admin-ui/internal/logging"
+)
+
+type Encrypt struct {
+	gcm cipher.AEAD
+
+	logger logging.LoggerInterface
```
shipperizer, line 20: for the sake of it, we should add the tracer

internal/encryption/encryption.go
```diff
@@ -0,0 +1,79 @@
...
+	ciphertext := e.gcm.Seal(nonce, nonce, payload, nil)
+	return hex.EncodeToString(ciphertext), nil
+}
+
+// Decrypt takes hex encoded string and returns the decrypted plain string
+func (e *Encrypt) Decrypt(hexData string) (string, error) {
+	encrypted, err := hex.DecodeString(hexData)
+	if err != nil {
+		e.logger.Errorf("error decoding hex encoded string, %v", err)
+		return "", fmt.Errorf("error decoding hex encoded string, %v", err)
```
shipperizer, line 40: same as above, provided it works

internal/encryption/encryption.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	noncePart, payloadPart := encrypted[:e.gcm.NonceSize()], encrypted[e.gcm.NonceSize():]
```
shipperizer, line 43: can this end up in an out of range panic?
might be better to check the length of slice

internal/encryption/encryption.go
```diff
@@ -0,0 +1,79 @@
...
+	decryptedData, err := e.gcm.Open(nil, noncePart, payloadPart, nil)
+	if err != nil {
+		e.logger.Errorf("error decrypting data: %v", err)
+		return "", fmt.Errorf("error decrypting data: %v", err)
+	}
+
+	return string(decryptedData), nil
+}
+
+var ioReadFull = io.ReadFull
```
shipperizer, line 53: is this needed for anything more than the casting?if we want to keep it can we put it top of the page?

internal/encryption/interfaces.go
```diff
@@ -0,0 +1,11 @@
+// Copyright 2024 Canonical Ltd.
+// SPDX-License-Identifier: AGPL-3.0
+
+package encryption
```
shipperizer, line 4: object in encryption seems to be used only in the authentication package, wouldn't it be better off under that instead of having a package on its own?

not a biggie, just don't wanna go nuclear with the amount of mini packages we have

internal/encryption/encryption.go
BarcoMasile: we need this only for testing purposes as apparently it's the only accepted way of controlling nested functions during unit tests. I kept it close to where it's used only for clarity but I can move it up top no problem

internal/encryption/encryption.go
BarcoMasile: Nonce is fixed at the advised 12 units. But yes, it would be better to check for data length. I'll change this and will have to adjust / enhance tests also

internal/encryption/encryption.go
BarcoMasile: yes it will work, just need to slightly refactor tests

internal/encryption/interfaces.go
BarcoMasile: Not a huge fan of mixing packages but I see what you mean, the language forces us already to have too many packages.
I'll merge it

nsklikas (commented):
suggestion: We could use https://github.com/gorilla/securecookie for signing and encrypting the cookies, to avoid having to deal with low level functions (e.g. https://github.com/canonical/lxd/blob/main/lxd/auth/oidc/oidc.go#L572)

BarcoMasile:
> suggestion: We could use https://github.com/gorilla/securecookie for signing and encrypting the cookies, to avoid having to deal with low level functions

So this seems nice, I have to admit. Thing is that switching to this right now would mean throw away 95% of the work done in this PR and starting over. I would probably consider this when/if we decide to support secret rotation for cookies encryption.
The good thing is that the std library has not been too complicated to use

shipperizer:
> > suggestion: We could use https://github.com/gorilla/securecookie for signing and encrypting the cookies, to avoid having to deal with low level functions
> 
> So this seems nice, I have to admit. Thing is that switching to this right now would mean throw away 95% of the work done in this PR and starting over. I would probably consider this when/if we decide to support secret rotation for cookies encryption.
> The good thing is that the std library has not been too complicated to use

let's add an issue on GitHub, if we can save ourselves from writing code we should (within reasons)

anyway lgtm

shipperizer (approved):

BarcoMasile:
> let's add an issue on GitHub, if we can save ourselves from writing code we should (within reasons)
> 
> anyway lgtm
Issue created: https://github.com/canonical/identity-platform-admin-ui/issues/330

nsklikas (approved):

PR #332, opened 2024-06-19
fix: ui uses relative base paths.

### LINKED ISSUE #317: Frontend pathing is broken
Frontend uses absolute paths for routing as well as for making calls to the backend API. This means that when the app is served behind a reverse proxy both the URLs used for routing and all calls to the backend do not work.

Bonus: Some api calls use an env var with an invalid  path (https://github.com/canonical/identity-platform-admin-ui/blob/main/ui/src/api/client.tsx#L10) and some use a correct path that is hardcoded (https://github.com/canonical/identity-platform-admin-ui/blob/main/ui/src/api/identities.tsx#L9). It looks like that same is true for the rebac components (https://github.com/canonical/rebac-admin/blob/main/src/api/authentication/authentication.ts#L43)

edlerd:
# Done
- fix to use relative base paths
  - in case /ui/ is found in the current page url, all links and api routes rely on the prefix of the current path before and including /ui/ as base url. If /ui/ is not found, fall back to / as the base path.
- streamline api base path definitions

Fixes #317
Fixes IAM-911
Fixes WD-12306

### DIFF
--- a/ui/src/App.tsx
+++ b/ui/src/App.tsx
@@ -11,6 +11,7 @@ import SchemaList from "pages/schemas/SchemaList";
 import Navigation from "components/Navigation";
 import Panels from "components/Panels";
 import useLocalStorage from "util/useLocalStorage";
+import { apiBasePath, basePath } from "util/basePaths";
 
 const App: FC = () => {
   // Store a user token that will be passed to the API using the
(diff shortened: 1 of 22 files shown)


ui/src/App.tsx
```diff
@@ -33,24 +34,24 @@ const App: FC = () => {
         <Suspense fallback={<Loader />}>
           <Routes>
             <Route
-              path="/"
+              path={basePath}
```
huwshimi, line 37: Instead of manipulating the routes, links and navigate calls you can set `basename` on the `<Router>` in index.tsx which will automatically make all other URLs relative to the base (see the last paragraph on this page: https://reactrouter.com/en/main/router-components/router).

ui/src/util/basePaths.ts
huwshimi, line 1: It looks like this file could do with some tests.

huwshimi (commented):
A couple of initial notes. I haven't done a QA yet.

BarcoMasile (dismissed):
I agree with Huw's comments.
I would be ok merging this since it fixes the linked issue, but at the same time I'd explore the use of `basename` prop directly on the Router object

ui/src/App.tsx
edlerd: Good idea, using the react router base path now.

ui/src/util/basePaths.ts
edlerd: Added test cases.

edlerd:
Adjusted to use the router property directly to keep this clean. Also added some test cases for the path calculation.

BarcoMasile (approved):

huwshimi (approved):
Thanks for the tests and the router basename improvement, that seems much cleaner.

One improvement in the future could be to create a central API URL map that prepends the basename when initialised.

PR #339, opened 2024-06-26
Add contextual tuples for global read and admin access

### LINKED ISSUE #290: Remove `global` from returned values when returning/listing Identities, Clients, Providers, Rules, Schemas, Roles and Groups.

nsklikas:
IAM-910

- Prefix the global object
- Use contextual tuples so that logic for global read and admin access is not persisted to openfga

One issue that we need to look into is that users will be able to create a role or a group called `__system__global` and it will collide with our tuples. One way around this is to add a check either on the roles/groups API or on the openfga module to prevent writes and deleted of objects that start with the system prefix. I can work on this on this PR or I can create an issue (I prefer the latter).

Once this PR is merged, we will no longer need to add the tuples with `<resource>:global` to openfga. I will remove them from the wiki.

I have tested it manually, but please take some time to test it yourselves as I might have missed some functionality.

Closes #290

### DIFF
--- a/internal/authorization/converters.go
+++ b/internal/authorization/converters.go
@@ -8,6 +8,7 @@ import (
 	"net/http"
 	"strings"
 
+	"github.com/canonical/identity-platform-admin-ui/internal/openfga"
 	"github.com/go-chi/chi/v5"
 )
 
@@ -25,11 +26,17 @@ const (
 	CAN_EDIT   = "can_edit"
 	CAN_CREATE = "can_create"
 	CAN_DELETE = "can_delete"
+
+	SYSTEM_OBJECT_PREFIX = "__system__"
+	// global is used in place of `*`, reason is OpenFGA has a special meaning for `*`
+	// see [public access](https://openfga.dev/docs/modeling/public-access)
+	GLOBAL_ACCESS_OBJECT_NAME = SYSTEM_OBJECT_PREFIX + "global"
 )
 
 type Permission struct {
-	Relation   string
-	ResourceID string
+	Relation         string
+	ResourceID       string
+	ContextualTuples []openfga.Tuple
 }
 
 func relation(r *http.Request) string {
@@ -55,15 +62,21 @@ func (c IdentityConverter) TypeName() string {
 
 func (c IdentityConverter) Map(r *http.Request) []Permission {
 	id := chi.URLParam(r, "id")
+	var contextualTuples []openfga.Tuple
 
 	if id == "" {
-		// global is used in place of `*`, reason is OpenFGA has a special meaning for `*`
-		// see [public access](https://openfga.dev/docs/modeling/public-access)
-		id = "global"
+		id = GLOBAL_ACCESS_OBJECT_NAME
+		// Add contextual tuples to enable all users to list resources and give admins access to all resources
+		globalResource := fmt.Sprintf("%s:%s", c.TypeName(), GLOBAL_ACCESS_OBJECT_NAME)
+		contextualTuples = append(
+			contextualTuples,
+			*openfga.NewTuple("user:*", CAN_VIEW, globalResource),
+			*openfga.NewTuple(ADMIN_OBJECT, PRIVILEGED_RELATION, globalResource),
+		)
 	}
 
 	return []Permission{
-		{Relation: relation(r), ResourceID: fmt.Sprintf("%s:%s", c.TypeName(), id)},
+		{Relation: relation(r), ResourceID: fmt.Sprintf("%s:%s", c.TypeName(), id), ContextualTuples: contextualTuples},
 	}
 }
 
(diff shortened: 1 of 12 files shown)


shipperizer (dismissed):

shipperizer:
need to update wiki docs and seed.sql as well

internal/authorization/converters.go
```diff
@@ -55,15 +62,20 @@ func (c IdentityConverter) TypeName() string {
...
 	if id == "" {
-		// global is used in place of `*`, reason is OpenFGA has a special meaning for `*`
-		// see [public access](https://openfga.dev/docs/modeling/public-access)
-		id = "global"
+		id = GLOBAL_ACCESS_OBJECT_NAME
+		// Add contextual tuples to enable all users to list resources and give admins access to all resources
+		contextualTuples = append(
+			contextualTuples,
+			*openfga.NewTuple("user:*", CAN_VIEW, fmt.Sprintf("%s:%s", c.TypeName(), GLOBAL_ACCESS_OBJECT_NAME)),
+			*openfga.NewTuple("privileged:superuser", ADMIN_OBJECT, fmt.Sprintf("%s:%s", c.TypeName(), GLOBAL_ACCESS_OBJECT_NAME)),
```
BarcoMasile, line 73: issue: ADMIN_OBJECT is equal to `privileged:user` (which identifies a type `privileged` with identifier `user`) which is not the name of an existing relation in the auth model. This should not work as it is.
I think it is supposed to be:


```go
*openfga.NewTuple(
    "privileged:superuser",
    "privileged",
    fmt.Sprintf("%s:%s", c.TypeName(), GLOBAL_ACCESS_OBJECT_NAME)
),
``` 

where `privileged` is the name of the relation of the type `identity` for example. Same goes for other types

I'll put a request changes for this just to alert, please double check

internal/authorization/converters.go
```diff
@@ -55,15 +62,20 @@ func (c IdentityConverter) TypeName() string {
...
 
 	if id == "" {
-		// global is used in place of `*`, reason is OpenFGA has a special meaning for `*`
-		// see [public access](https://openfga.dev/docs/modeling/public-access)
-		id = "global"
+		id = GLOBAL_ACCESS_OBJECT_NAME
+		// Add contextual tuples to enable all users to list resources and give admins access to all resources
+		contextualTuples = append(
+			contextualTuples,
+			*openfga.NewTuple("user:*", CAN_VIEW, fmt.Sprintf("%s:%s", c.TypeName(), GLOBAL_ACCESS_OBJECT_NAME)),
```
BarcoMasile, line 72: nitpick: why not saving the `Sprintf` result in a variable to reuse in these two tuples, instead of computing it twice? To push it a step further, we could have contants for these values since typename and global object name will remain the same

This nitpick applies to all other changed lines with code like this in the pr

BarcoMasile (changes requested):

internal/authorization/converters.go
shipperizer: in all fairness this is not needed at all, the link between the `privileged` superuser and the objects is done at creation,  so only one that needs to be provided is the `user:*`

i'll try to set this up via openfga diagram to explain better 

```suggestion
```

shipperizer:
Screenshot from 2024-06-27 11-26-10

ok went a bit deeper to recall what happens under the hood

so from an authorization perspective what we need to check when we do a `GET /api/v0/<resources>` is to simply check if we have the `can_view` permission on `<resource>:<global_tag>`

from this perspective all is the same across different `backends`,be it OpenFGA or Hydra/Kratos etc


where it differs between resources is when we ask for the resources to the underlying backend, so for Hydra/Kratos/Oathkeeper we get all due to issues with pagination and filtering

for OpenFGA resources (`role` and `group`) instead we are able to show only what we are matched with via the `can_view`, which is different from the `assignee`

in the latter case having the contextual tuple might be beneficial if we wanted to say "see all roles, no matter what`, not sure we want that at this point though

internal/authorization/converters.go
shipperizer: my mistake, `privileged` contextual tuple needs to be there for other action types (like create or edit) and allows users in provided they are superusers

shipperizer:
> Screenshot from 2024-06-27 11-26-10
> 
> ok went a bit deeper to recall what happens under the hood
> 
> so from an authorization perspective what we need to check when we do a `GET /api/v0/<resources>` is to simply check if we have the `can_view` permission on `<resource>:<global_tag>`
> 
> from this perspective all is the same across different `backends`,be it OpenFGA or Hydra/Kratos etc
> 
> where it differs between resources is when we ask for the resources to the underlying backend, so for Hydra/Kratos/Oathkeeper we get all due to issues with pagination and filtering
> 
> for OpenFGA resources (`role` and `group`) instead we are able to show only what we are matched with via the `can_view`, which is different from the `assignee`
> 
> in the latter case having the contextual tuple might be beneficial if we wanted to say "see all roles, no matter what`, not sure we want that at this point though

after chat, all is clear, my bad

BarcoMasile (approved):

shipperizer (approved):

PR #341, opened 2024-06-27
IAM 887 Add `/auth/me` endpoint to retrieve info about currently logged user

BarcoMasile:
## Description
This PR builds on top of #338 , so it will need a rebase after that one is merged.
It introduces a simple `me` handler to allow clients to retrieve info about the currently logged user.
This is useful for frontend applications which can rely on this endpoint to know who is the user that is currently logged in.

## Changes
- `isAllowListed` method now relies on an exact match for it's matching strategy, leaving the previous prefix based matching behind
- `nonce` field is now ignored when serializing the `Principal` struct

### DIFF
--- a/pkg/authentication/middlewares.go
+++ b/pkg/authentication/middlewares.go
@@ -40,12 +40,8 @@ func (m *Middleware) SetAllowListedEndpoints(endpointsPrefixes ...string) {
 
 func (m *Middleware) isAllowListed(r *http.Request) bool {
 	endpoint := r.URL.Path
-	for prefix, _ := range m.allowListedEndpoints {
-		if strings.HasPrefix(endpoint, prefix) {
-			return true
-		}
-	}
-	return false
+	_, ok := m.allowListedEndpoints[endpoint]
+	return ok
 }
 
 func (m *Middleware) OAuth2AuthenticationChain() []func(http.Handler) http.Handler {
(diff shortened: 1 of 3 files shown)


natalian98 (approved):

pkg/authentication/middlewares.go
```diff
@@ -40,12 +40,8 @@ func (m *Middleware) SetAllowListedEndpoints(endpointsPrefixes ...string) {
 
 func (m *Middleware) isAllowListed(r *http.Request) bool {
 	endpoint := r.URL.Path
-	for prefix, _ := range m.allowListedEndpoints {
-		if strings.HasPrefix(endpoint, prefix) {
-			return true
-		}
-	}
-	return false
+	_, ok := m.allowListedEndpoints[endpoint]
```
shipperizer, line 43: what happens in the case of the slice being not initialized?does `range` deal with the nil pointer check?if not happy to swap otherwise we might need to check for the remote edge case

BarcoMasile: This cannot happen since the construction funtion takes care of initializing the map. Plus, with `isAllowListed` being a "private" method we should be good to go.
But I'm open to refactor this

shipperizer: nah that's good enough for me then

shipperizer (approved):

PR #345, opened 2024-07-03
IAM 889 Implement logout feature

BarcoMasile:
## Description
Adds the logout handler `/api/v0/auth/logout` which does two things mainly
- deletes the hydra session (in case the Principal is a browser user, otherwise no session info is availble for CLI)
- revokes the refresh token (if present, access token otherwise) so they cannot be used anymore

Of course it also deletes the encrypted auth cookies.

Hydra admin url is used for the revoke session API invocation, here we rely on hydra implementation of this so I made explicit reference to hydra url since there is no need to keep it generic. In case we ever change from Hydra to another OIDC provider, we'll have to switch the logout implementation anyway.

This PR also fixes cookie deletion for the cookie manager, since a bug was present and it slipped in the PR it introduced it.

**NB**: The revoke token API invocation is based on the fact that the token endpoint auth method is set to client-secret-basic at client creation (against Hydra), and this is set in the admin-ui-operator right [here](https://github.com/canonical/identity-platform-admin-ui-operator/blob/cbef94404f00055034412e34c357faa6758b2cd9/lib/charms/hydra/v0/oauth.py#L271).
If we ever change client-authentication to `client-secret-post`, this will mean that we will have to add client id and secret to the revoke token request body. I don't see a reason why we would do that, but the info is here for posterity.

This PR will benefit from solving issue #337 , which will be worked on right after this.

### DIFF
--- a/pkg/authentication/oidc.go
+++ b/pkg/authentication/oidc.go
@@ -57,16 +57,29 @@ func PrincipalFromContext(ctx context.Context) *Principal {
 		return nil
 	}
 
-	value := ctx.Value(PrincipalContextKey)
-	if value == nil {
-		return nil
+	if value, ok := ctx.Value(PrincipalContextKey).(*Principal); ok {
+		return value
 	}
 
-	return value.(*Principal)
+	return nil
 }
 
 func OtelHTTPClientContext(ctx context.Context) context.Context {
 	return context.WithValue(ctx, oauth2.HTTPClient, otelHTTPClient)
+}
+
+func HTTPClientFromContext(ctx context.Context) *http.Client {
+	client := http.DefaultClient
+
+	if ctx == nil {
+		return client
+	}
+
+	if c, ok := ctx.Value(oauth2.HTTPClient).(*http.Client); ok {
+		client = c
+	}
+
+	return client
 }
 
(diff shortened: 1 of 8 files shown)


pkg/authentication/oidc.go
```diff
@@ -119,6 +133,110 @@ func (o *OAuth2Context) RefreshToken(ctx context.Context, rawRefreshToken string
...
+	err = json.NewEncoder(&buf).Encode(&revokeTokenRequest{Token: token})
+	if err != nil {
+		return nil, err
+	}
+
+	req.SetBasicAuth(o.client.ClientID, o.client.ClientSecret)
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+	req.Body = io.NopCloser(&buf)
+
+	return req, nil
```
shipperizer, line 237: why are we not using the hydra sdk for this?

BarcoMasile: Clarified offline, will proceed and refactor implementation and tests. Thanks

BarcoMasile: Done

shipperizer (approved):

PR #346, opened 2024-07-05
Improve validation error messages

nsklikas:
IAM-931

Refactor the validation error response messages
image

### DIFF
--- a/internal/validation/types.go
+++ b/internal/validation/types.go
@@ -40,15 +40,22 @@ func buildErrorData(errors validator.ValidationErrors) map[string][]string {
 	for _, e := range errors {
 		field := e.Field()
 
+		var errMsg string
+		switch e.Tag() {
+		case "required":
+			errMsg = fmt.Sprintf("Missing required field '%s'", field)
+		case "oneof", "httpmethod":
+			errMsg = fmt.Sprintf("Field '%s' must be one of '%s'", field, e.Param())
+		default:
+			errMsg = e.Error()
+		}
+
 		failures, ok := failedValidations[field]
 		if !ok {
 			failedValidations[field] = make([]string, 0)
 		}
 
-		failures = append(
-			failures,
-			fmt.Sprintf("value '%s' fails validation of type `%s`", e.Value(), e.Tag()),
-		)
+		failures = append(failures, errMsg)
 		failedValidations[field] = failures
 	}
 
--- a/pkg/clients/validation.go
+++ b/pkg/clients/validation.go
@@ -6,12 +6,14 @@ package clients
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"strings"
 
 	"github.com/go-playground/validator/v10"
 	client "github.com/ory/hydra-client-go/v2"
 
+	"github.com/canonical/identity-platform-admin-ui/internal/logging"
 	"github.com/canonical/identity-platform-admin-ui/internal/validation"
 )
 
@@ -32,6 +34,8 @@ var (
 type PayloadValidator struct {
 	apiKey    string
 	validator *validator.Validate
+
+	logger logging.LoggerInterface
 }
 
 func (p *PayloadValidator) setupValidator() {
@@ -49,7 +53,9 @@ func (p *PayloadValidator) Validate(ctx context.Context, method, endpoint string
 	if p.isCreateClient(method, endpoint) || p.isUpdateClient(method, endpoint) {
 		clientRequest := new(client.OAuth2Client)
 		if err := json.Unmarshal(body, clientRequest); err != nil {
-			return ctx, nil, err
+			// TODO(nsklikas): evaluate usage of logger
+			p.logger.Error("Json parsing error: ", err)
+			return ctx, nil, fmt.Errorf("failed to parse JSON body")
 		}
 
 		err = p.validator.Struct(clientRequest)
@@ -76,9 +82,10 @@ func (p *PayloadValidator) isUpdateClient(method string, endpoint string) bool {
 	return method == http.MethodPut && strings.HasPrefix(endpoint, "/")
 }
 
-func NewClientsPayloadValidator(apiKey string) *PayloadValidator {
+func NewClientsPayloadValidator(apiKey string, logger logging.LoggerInterface) *PayloadValidator {
 	p := new(PayloadValidator)
 	p.apiKey = apiKey
+	p.logger = logger
 	p.validator = validation.NewValidator()
 
 	p.setupValidator()
--- a/pkg/groups/validation.go
+++ b/pkg/groups/validation.go
@@ -6,18 +6,24 @@ package groups
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 	"net/http"
 	"strings"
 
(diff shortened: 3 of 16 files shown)


pkg/identities/validation.go
```diff
@@ -50,7 +54,8 @@ func (p *PayloadValidator) Validate(ctx context.Context, method, endpoint string
 	if p.isCreateIdentity(method, endpoint) {
 		createIdentity := new(CreateIdentityRequest)
 		if err := json.Unmarshal(body, createIdentity); err != nil {
-			return ctx, nil, err
+			p.logger.Info("Json parsing error: ", err)
```
BarcoMasile, line 57: This should be `p.logger.Error(...)`
same below

internal/validation/types.go
```diff
@@ -31,25 +31,26 @@ func NewValidationError(msg string, errors validator.ValidationErrors) *types.Re
...
 
-func buildErrorData(errors validator.ValidationErrors) map[string][]string {
+func buildErrorData(errors validator.ValidationErrors) []string {
 	if errors == nil {
 		return nil
 	}
 
-	failedValidations := make(map[string][]string)
-	for _, e := range errors {
+	failedValidations := make([]string, len(errors))
```
BarcoMasile, line 39: issue: the payload validation with specific errors was done for the UI to consume so the user could be presented with field-specific validation errors to act upon.
Switching from a map to an array would invalidate this benefit, making the validation errors useless, since the UI would not be able to grab field-specific errors.
I suggest you stick to the map structure

BarcoMasile (commented):
LGTM apart from the switch from map to slice, which I commented about.
Nitpick: since all validators are only using the logger, I would remove the tracer and monitor.

pkg/identities/validation.go
nsklikas: Most of the time it will happen because someone sent a malformed json, so imho it is not worth logging it as an error. If you believe this will add value as an error, I have no problem to change it back.

pkg/rules/validation.go
```diff
@@ -61,7 +65,8 @@ func (p *PayloadValidator) Validate(ctx context.Context, method, endpoint string
 	if p.isCreateOrUpdateRule(method, endpoint) {
 		ruleRequest := new(oathkeeper.Rule)
 		if err := json.Unmarshal(body, ruleRequest); err != nil {
-			return ctx, nil, err
+			p.logger.Info("Json parsing error: ", err)
+			return ctx, nil, fmt.Errorf("failed to parse JSON body")
```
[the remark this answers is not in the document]
nsklikas: I removed it because I thought that most of the time the go error is not user friendly, so it isn't very helpful to expose to the user. No problem with adding it back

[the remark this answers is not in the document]
BarcoMasile: that's ok as it is then, thanks

BarcoMasile (dismissed):

pkg/clients/validation.go
```diff
@@ -49,7 +53,8 @@ func (p *PayloadValidator) Validate(ctx context.Context, method, endpoint string
 	if p.isCreateClient(method, endpoint) || p.isUpdateClient(method, endpoint) {
 		clientRequest := new(client.OAuth2Client)
 		if err := json.Unmarshal(body, clientRequest); err != nil {
-			return ctx, nil, err
+			p.logger.Error("Json parsing error: ", err)
```
shipperizer, line 56: are we using the `fmt.Errorf("failed to parse JSON body")` to bubble it up to the frontend?
if not can we drop the logger from here and simply enrich the error with `err`?

```suggestion
return ctx, nil, fmt.Errorf("failed to parse JSON body: %s", err)
```

pkg/groups/validation.go
```diff
@@ -35,7 +41,8 @@ func (p *PayloadValidator) Validate(ctx context.Context, method, endpoint string
 	if p.isCreateGroup(method, endpoint) {
```
wood-push-melon, line 41: Just a **NONBLOCKING** comment: (understand this is out of this PR's scope.)

These `if` blocks smell a bit risky if some endpoints are added or deleted (open/closed principle). I guess it will be improved in the future?

nsklikas: I agree, not sure if there is a plan for that (I also don't like how it's a bunch of `if`s, but in reality it is more like a `switch` or `if/else` statement)

internal/validation/types.go
```diff
@@ -40,15 +40,22 @@ func buildErrorData(errors validator.ValidationErrors) map[string][]string {
...
+		case "oneof", "httpmethod":
+			errMsg = fmt.Sprintf("Field '%s' must be one of '%s'", field, e.Param())
+		default:
+			errMsg = e.Error()
+		}
+
 		failures, ok := failedValidations[field]
 		if !ok {
 			failedValidations[field] = make([]string, 0)
 		}
```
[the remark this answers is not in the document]
nsklikas: Not sure if I understand correctly, we check if it exists because the first time we insert an element we need to initialize the string slice.

pkg/clients/validation.go
nsklikas: Yes, my problem was that previously the message was pretty useless for another service or for a user (it was something like `Unable to parse into Go struct`). I prefer this, but no problem with reverting it.

pkg/groups/validation.go
BarcoMasile: Unfortunately there is no way to generalize the validation of the payload from the controller logic that needs to handle it, so when a new endpoint gets added then necessarily you need to add validation for it, otherwise you just can't validate that payload
Also, notice that same types are validated differently based on which handlers they are for (for example POST Group or PUT Group, different validation).

`if`s are usually not nice to look at but branching in this case is more dynamic and based on more than one value, and the alternative would be to hide the "ifs" with an OOP approach, that's why this was kept as simple as possible.

I'm sure we'll focus on improving things after deadlines are met.

BarcoMasile: @wood-push-melon open an issue with your suggestion so we can track it and keep it in mind for improvements!

internal/validation/types.go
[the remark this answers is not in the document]
wood-push-melon: Ok, I remember Go does automatically initialization for the value. So we cannot directly do something like this?

```
failedValidations[field] = append(failedValidations[field], errMsg)
```

Just a minor thing, feel free to ignore this.

pkg/groups/validation.go
wood-push-melon: gotcha. will do. thanks.

internal/validation/types.go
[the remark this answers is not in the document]
nsklikas: not if they are a slice. You need to allocate the necessary memory for slices, maps and structs

pkg/clients/validation.go
shipperizer: not a biggie in the great scheme of things, do me a favour, add a `TODO` saying to review usage of logger and let s move on

shipperizer (approved):

BarcoMasile (approved):

shipperizer (approved):

PR #348, opened 2024-07-08
IAM 933 - Fix UI serving and implement `next` param passing for FE deep redirection

### LINKED ISSUE #337: After login/logout flow completed, handler should redirect to frontend correctly, also allowing FE to show the user the same page that initiated login/logout

BarcoMasile:
## Description
Two things come with this PR
- a small fix to the ui serving handlers and semplification of the implementation
- a feature which allows the frontend to pass a `next` query param with the login GET request, which allows the FE to get that value back **after** login completes so it can use/interpret it (this should be used by the FE to know where the login request originated from in order to bring the user back to that UI).
The `next` parameter does not need to be a relative url or anything similar, it's just parsed as a string and it is returned as it is.
The UI can use it in whichever way it wants.

This closes #337

### DIFF
--- a/internal/config/specs.go
+++ b/internal/config/specs.go
@@ -12,7 +12,8 @@ type EnvSpec struct {
 	LogLevel string `envconfig:"log_level" default:"error"`
 	LogFile  string `envconfig:"log_file" default:"log.txt"`
 
-	Port int `envconfig:"port" default:"8080"`
+	Port        int    `envconfig:"port" default:"8080"`
+	ContextPath string `envconfig:"context_path" default:"/"`
 
 	Debug bool `envconfig:"debug" default:"false"`
 
--- a/internal/config/specs.go
+++ b/internal/config/specs.go
@@ -12,8 +12,7 @@ type EnvSpec struct {
 	LogLevel string `envconfig:"log_level" default:"error"`
 	LogFile  string `envconfig:"log_file" default:"log.txt"`
 
-	Port    int    `envconfig:"port" default:"8080"`
-	BaseURL string `envconfig:"base_url" default:""`
+	Port int `envconfig:"port" default:"8080"`
 
 	Debug bool `envconfig:"debug" default:"false"`
 
--- a/pkg/authentication/cookies.go
+++ b/pkg/authentication/cookies.go
@@ -15,6 +15,7 @@ const (
 	authCookiePath    = "/api/v0/auth/callback"
 	nonceCookieName   = "nonce"
 	stateCookieName   = "state"
+	nextToCookieName  = "nextTo"
 	itCookieName      = "id-token"
 	atCookieName      = "access-token"
 	rtCookieName      = "refresh-token"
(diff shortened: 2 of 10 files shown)


pkg/authentication/handlers.go
```diff
@@ -87,6 +88,10 @@ func (a *API) handleLogin(w http.ResponseWriter, r *http.Request) {
 	// add the Otel HTTP Client
 	r = r.WithContext(OtelHTTPClientContext(r.Context()))
 
+	if nextTo := r.URL.Query().Get("next"); nextTo != "" {
+		a.cookieManager.SetNextToCookie(w, nextTo)
+	}
```
nsklikas, line 93: We need to validate the value of `next`. We should allow redirects only to our domain, otherwise anyone will be able to get the session cookies

pkg/authentication/handlers.go
```diff
@@ -158,7 +163,16 @@ func (a *API) handleCallback(w http.ResponseWriter, r *http.Request) {
 	a.cookieManager.SetAccessTokenCookie(w, oauth2Token.AccessToken)
 	a.cookieManager.SetRefreshTokenCookie(w, oauth2Token.RefreshToken)
 
-	http.Redirect(w, r, ui.UIPrefix, http.StatusFound)
+	redirectURL, _ := url.ParseRequestURI(ui.UIPrefix)
```
nsklikas, line 166: Will this work when running under a path? (I think the `base_url` is needed for this)

BarcoMasile: Right, I'll go and add a `ContextPath` spec to handle this. Thank you!

pkg/authentication/cookies.go
```diff
@@ -135,7 +148,6 @@ func (a *AuthCookieManager) clearCookie(w http.ResponseWriter, name string, path
 func (a *AuthCookieManager) getCookie(r *http.Request, name string) string {
 	cookie, err := r.Cookie(name)
 	if err != nil {
-		a.logger.Errorf("can't get cookie %s, %v", name, err)
 		return ""
```
shipperizer, line 151: if we are not returning nor logging the errors do we need the if check at all? shouldn't we simply check if cookie is nil for example?

otherwise worth returning or logging the error

shipperizer (commented):
let's make sure to test this

pkg/authentication/cookies.go
BarcoMasile: The error is always "cookie not found", which we don't care, we only need the value if present
Empty string is enough to tell us no cookie was found

internal/config/specs.go
```diff
@@ -12,8 +12,8 @@ type EnvSpec struct {
 	LogLevel string `envconfig:"log_level" default:"error"`
 	LogFile  string `envconfig:"log_file" default:"log.txt"`
 
-	Port    int    `envconfig:"port" default:"8080"`
-	BaseURL string `envconfig:"base_url" default:""`
+	Port        int    `envconfig:"port" default:"8080"`
+	ContextPath string `envconfig:"context_path" default:"/"`
```
nsklikas, line 16: Can you update the docs with this chage

nsklikas (commented):
One final thing for me

nsklikas (approved):

shipperizer (approved):

PR #349, opened 2024-07-09
Add `can_view` entitlement for each resource creation

### LINKED ISSUE #298: Add `can_view` entitlement for each resource creation, with current user as OpenFGA tuple user
Right now when creating a resource, the user doing it won't have entitlements to actually list that resource (seeing it appear in the list returned by the `GET /{resource}` APIs.
Same goes for the `GET /{resource}/{id}` APIs.
To fix this, we need to also add a tuple for `can_view` entitlements like what's been done in #295 .

nsklikas:
IAM-866

If we agree with this I will refactor all APIs. My main purpose with this approach is to keep the authz logic away from each API and to contain it in the `authorization` package.

Also fixes some of the logic for initializing the authorizer (@shipperizer @BarcoMasile please have a look that I don't break something in `serve.go` and `router.go`.

Closes #298

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -90,14 +90,8 @@ func serve() {
 				logger,
 			),
 		),
-		// default to noop client for authorization
-		openfga.NewNoopClient(tracer, monitor, logger),
+		nil,
 	)
-
-	if specs.AuthorizationEnabled {
-		logger.Info("Authorization is enabled")
-		externalConfig.SetAuthorizer(externalConfig.OpenFGA())
-	}
 
 	k8sCoreV1, err := k8s.NewCoreV1Client(specs.KubeconfigFile)
 
@@ -132,10 +126,21 @@ func serve() {
 			monitor,
 			logger,
 		)
+		logger.Info("Authorization is enabled")
+		externalConfig.SetAuthorizer(authorizer)
 
 		if authorizer.ValidateModel(context.Background()) != nil {
 			panic("Invalid authorization model provided")
 		}
+	} else {
+		authorizer := authorization.NewAuthorizer(
+			openfga.NewNoopClient(tracer, monitor, logger),
+			tracer,
+			monitor,
+			logger,
+		)
+		logger.Info("Using noop authorizer")
+		externalConfig.SetAuthorizer(authorizer)
 	}
 
 	oauth2Config := authentication.NewAuthenticationConfig(
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -119,9 +119,13 @@ func serve() {
 		DistFS: distFS,
 	}
 
+	wpool := pool.NewWorkerPool(specs.OpenFGAWorkersTotal, tracer, monitor, logger)
+	defer wpool.Stop()
+
 	if specs.AuthorizationEnabled {
 		authorizer := authorization.NewAuthorizer(
 			externalConfig.OpenFGA(),
+			wpool,
 			tracer,
 			monitor,
 			logger,
@@ -135,6 +139,7 @@ func serve() {
 	} else {
 		authorizer := authorization.NewAuthorizer(
 			openfga.NewNoopClient(tracer, monitor, logger),
+			wpool,
 			tracer,
 			monitor,
 			logger,
@@ -161,9 +166,6 @@ func serve() {
 	ollyConfig := web.NewO11yConfig(tracer, monitor, logger)
 
 	routerConfig := web.NewRouterConfig(specs.ContextPath, specs.PayloadValidationEnabled, idpConfig, schemasConfig, rulesConfig, uiConfig, externalConfig, oauth2Config, ollyConfig)
-
-	wpool := pool.NewWorkerPool(specs.OpenFGAWorkersTotal, tracer, monitor, logger)
-	defer wpool.Stop()
 
 	router := web.NewRouter(routerConfig, wpool)
 
--- a/internal/authorization/authorization.go
+++ b/internal/authorization/authorization.go
@@ -10,6 +10,7 @@ import (
 	"github.com/canonical/identity-platform-admin-ui/internal/logging"
 	"github.com/canonical/identity-platform-admin-ui/internal/monitoring"
 	"github.com/canonical/identity-platform-admin-ui/internal/openfga"
+	"github.com/canonical/identity-platform-admin-ui/internal/pool"
 	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
 )
 
@@ -17,6 +18,8 @@ var ErrInvalidAuthModel = fmt.Errorf("Invalid authorization model schema")
 
 type Authorizer struct {
 	client AuthzClientInterface
+
+	wpool pool.WorkerPoolInterface
 
 	tracer  tracing.TracingInterface
 	monitor monitoring.MonitorInterface
@@ -71,9 +74,10 @@ func (a *Authorizer) ValidateModel(ctx context.Context) error {
 	return nil
 }
 
-func NewAuthorizer(client AuthzClientInterface, tracer tracing.TracingInterface, monitor monitoring.MonitorInterface, logger logging.LoggerInterface) *Authorizer {
+func NewAuthorizer(client AuthzClientInterface, wpool pool.WorkerPoolInterface, tracer tracing.TracingInterface, monitor monitoring.MonitorInterface, logger logging.LoggerInterface) *Authorizer {
 	authorizer := new(Authorizer)
 	authorizer.client = client
+	authorizer.wpool = wpool
 	authorizer.tracer = tracer
 	authorizer.monitor = monitor
 	authorizer.logger = logger
--- a/internal/authorization/resource_manager.go
+++ b/internal/authorization/resource_manager.go
@@ -1,0 +1,79 @@
+package authorization
+
+import (
+	"context"
+)
+
+func (a *Authorizer) createTuple(ctx context.Context, resourceID string) error {
+	user, ok := ctx.Value(USER_CTX).(*User)
+	if !ok {
+		// Should we panic?
+		return nil
+	}
+	go func() {
+		err := a.client.WriteTuple(context.Background(), "user:"+user.ID, CAN_VIEW, resourceID)
+		if err != nil {
+			a.logger.Errorf("Failed to create authorization tuple: %s", err)
+		}
+	}()
+	return nil
+}
+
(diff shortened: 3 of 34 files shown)


internal/authorization/authorization.go
```diff
@@ -32,6 +37,13 @@ func (a *Authorizer) Check(ctx context.Context, user string, relation string, ob
 	return a.client.Check(ctx, user, relation, object, contextualTuples...)
 }
 
+func (a *Authorizer) WriteTuple(ctx context.Context, user string, relation string, object string) error {
+	ctx, span := a.tracer.Start(ctx, "authorization.Authorizer.WriteTuple")
+	defer span.End()
+
+	return a.client.WriteTuple(ctx, user, relation, object)
```
shipperizer, line 44: I d use the WriteTuples on the low level object of OpenFGA client, just in case we need more tuples.to be set

at the same time this is kinda blocking and should not be imho, so needs to be offloaded to a background worker/object

internal/authorization/helpers.go
```diff
@@ -0,0 +1,19 @@
...
+func (a *ClientAuthorizer) ClientResource(clientID string) string {
+	return "client:" + clientID
+}
+
+func (a *ClientAuthorizer) CreateClient(ctx context.Context, clientID string) error {
+	user := ctx.Value(USER_CTX)
+	if user == nil {
+		return nil
+	}
+	return a.client.WriteTuple(ctx, "user:"+user.(*User).ID, CAN_VIEW, a.ClientResource(clientID))
```
BarcoMasile, line 18: suggestion: You lack the check for the typing to `*User` being successful.
Taking a look at [this](https://github.com/canonical/identity-platform-admin-ui/blob/main/pkg/authentication/oidc.go#L64-L66), you could improve this and add the check to make sure that `user.(*User)` goes well by doing:

```suggestion
	if user, ok := ctx.Value(USER_CTX).(*User); !ok {
		return nil
	}
	
	return a.client.WriteTuple(ctx, "user:"+user.ID, CAN_VIEW, a.ClientResource(clientID))
```

This way when you use the `user` object you know 100% it is a legitimate `*User` type.

internal/authorization/helpers.go
```diff
@@ -0,0 +1,19 @@
+package authorization
+
+import (
+	"context"
+)
+
+type ClientAuthorizer = authorizer
```
BarcoMasile, line 7: I'm not a big fan of this approach for two reasons:
- if you use it as a new object you need to initiate it, but then you would initiate an entire authorizer just for the helper methods
- if you use this starting from an existing object, to have these method available then you would need to cast the existing object to `ClientAuthorizer`

My opinion on this is that it makes things unnecessarily more complicated than they need to be.
For example the `ClientResource` method could be an exported utility function since it's very useful
Meanwhile the `CreateClient` seems to belong the main Authorizer struct.

internal/authorization/resource_manager.go
```diff
@@ -0,0 +1,79 @@
...
+	if !ok {
+		// Should we panic?
+		return nil
+	}
+	go func() {
+		err := a.client.WriteTuple(context.Background(), "user:"+user.ID, CAN_VIEW, resourceID)
+		if err != nil {
+			a.logger.Errorf("Failed to create authorization tuple: %s", err)
+		}
+	}()
```
shipperizer, line 18: we can't do this, have a look at how we use the [pool](https://github.com/canonical/identity-platform-admin-ui/blob/main/internal/pool/pool.go) in [here](https://github.com/canonical/identity-platform-admin-ui/blob/main/pkg/roles/service.go#L286-L289)

reason is this is directly affected by the number of requests on a Create operation, which would break the limit of file descriptors if put under stress

internal/authorization/resource_manager.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"context"
+)
+
+func (a *Authorizer) createTuple(ctx context.Context, resourceID string) error {
+	user, ok := ctx.Value(USER_CTX).(*User)
+	if !ok {
+		// Should we panic?
+		return nil
```
shipperizer, line 11: return an error `fmt.Errorf("user not authenticated or requests context broken, cannot perform operation")`

internal/authorization/resource_manager.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+	go func() {
+		err := a.client.DeleteTuple(context.Background(), "user:"+user.ID, CAN_VIEW, resourceID)
+		if err != nil {
+			a.logger.Errorf("Failed to create authorization tuple: %s", err)
+		}
+	}()
+	return nil
```
shipperizer, line 34: same as above

pkg/web/interfaces.go
```diff
@@ -30,6 +30,21 @@ type OpenFGAClientInterface interface {
...
+	CreateIdentity(context.Context, string) error
+	DeleteIdentity(context.Context, string) error
+	CreateProvider(context.Context, string) error
+	DeleteProvider(context.Context, string) error
+	CreateRule(context.Context, string) error
+	DeleteRule(context.Context, string) error
+	CreateSchema(context.Context, string) error
+	DeleteSchema(context.Context, string) error
+}
+
```
shipperizer, line 47: ideally we move this into each package where each `AuthorizerClientInterface` implements a subset of these being used

aka `identities` pkg will have

```
type AuthorizerClientInterface interface {
	CreateIdentity(context.Context, string) error
	DeleteIdentity(context.Context, string) error
}
```

while  `idps` pkg will have

```
type AuthorizerClientInterface interface {
	CreateProvider(context.Context, string) error
	DeleteProvider(context.Context, string) error
}
```

in here we can simply pass the pointer to the Authorizer

pkg/web/interfaces.go
```diff
@@ -40,5 +55,5 @@ type ExternalClientsConfigInterface interface {
 	HydraAdmin() *ih.Client
 	KratosAdmin() *ik.Client
 	OpenFGA() OpenFGAClientInterface
-	Authorizer() OpenFGAClientInterface
+	Authorizer() AuthorizerClientInterface
```
shipperizer, line 58:
```suggestion
	Authorizer() *Authorizer
```

pkg/web/interfaces.go
nsklikas: That's already implemented (eg https://github.com/canonical/identity-platform-admin-ui/pull/349/files#diff-ecd1991c6a0e8a942b9f3669dff799d8588c817d694555198c6caa0f9986ea58).

internal/authorization/resource_manager.go
nsklikas: I switched it to use the pool. It feels like the pool was not supposed to work like this (async jobs that we don't wish to wait for), but I guess it works as a temporary workaround.

pkg/web/interfaces.go
nsklikas: fixed

cmd/serve.go
```diff
@@ -119,9 +119,13 @@ func serve() {
 		DistFS: distFS,
 	}
 
+	wpool := pool.NewWorkerPool(specs.OpenFGAWorkersTotal, tracer, monitor, logger)
```
BarcoMasile, line 122: worker pools are used to have control over the amount of resources we use.
The other worker pool still interacts with OFGA. My suggestion is to reuse that one (or delete that one and reuse this) since we always hit the same OpenFGA

internal/authorization/resource_manager.go
```diff
@@ -0,0 +1,80 @@
...
+	user, ok := ctx.Value(USER_CTX).(*User)
+	if !ok {
+		return ErrNoUserInContext
+	}
+
+	_, err := a.wpool.Submit(
+		a.client.WriteTuple(context.Background(), "user:"+user.ID, CAN_VIEW, resourceID),
+		nil,
+		nil,
+	)
```
BarcoMasile, line 34: issue: the first parameter to the `Submit` must be a callable, either a 
`func()` or a `func() any`

I'm pretty sure this doesn't work in parallel but just invokes the write tuple right here.
This is supposed to be like

```suggestion
	_, err := a.wpool.Submit(
	        func() any {
	            return a.client.WriteTuple(context.Background(), "user:"+user.ID, CAN_VIEW, resourceID)
	        },
		nil,
		nil,
	)
```

BarcoMasile (changes requested):

internal/authorization/resource_manager.go
shipperizer: at the same time the function should inspect the result and log if error

cmd/serve.go
nsklikas: Even though I think I agree with @shipperizer, this is the same pool that we use for all our other components

internal/authorization/resource_manager.go
nsklikas: thanks, fixed

internal/authorization/resource_manager.go
```diff
@@ -0,0 +1,90 @@
...
+
+func (a *Authorizer) createTuple(ctx context.Context, resourceID string) error {
+	user, ok := ctx.Value(USER_CTX).(*User)
+	if !ok {
+		return ErrNoUserInContext
+	}
+
+	_, err := a.wpool.Submit(
+		func() any {
+			return a.client.WriteTuple(context.Background(), "user:"+user.ID, CAN_VIEW, resourceID)
```
shipperizer, line 18:
```suggestion
			if err := a.client.WriteTuple(context.Background(), "user:"+user.ID, CAN_VIEW, resourceID); err != nil {
			    a.logger.Error(err.Error())
			    return err
			}
			return nil
```

cmd/serve.go
BarcoMasile: Thanks, sorry I mistook that

BarcoMasile (dismissed):
LGTM
Only thing I want to point out, but I'm not letting this stop the merging:

Only thing I'm not a fan of is naming of the methods for the authz interface.
I would suggest something more like
`UpdateEntitlementsFor<API-Operation>` or `SetPermissionsFor<API-Operation>`. This way it's 100% clear what it does even to who reads the code and doesn't have a full grasp on what we do on the backend.

Example:
- `UpdateEntitlementsForCreateClient`/`SetPermissionsForCreateClient`

nsklikas:
I agree, that is clearer. I updated the name, should be fine now

BarcoMasile (approved):

shipperizer (approved):

BarcoMasile (approved):

PR #359, opened 2024-07-16
Set cookie path to /

nsklikas:
Needed for when the api is served under a subpath


BarcoMasile (approved):
Good one, thanks for this.
I think it's safe to merge if the tests pass

natalian98 (approved):

PR #360, opened 2024-07-17
IAM-891 link authenticated users to authorization model

BarcoMasile:
## Description
This PR links authenticated principals to the authorization model in OpenFGA.
This means that either a ServicePrincipal (CLI use case) or a UserPrincipal (browser use case) will be "mapped" to a OpenFGA user of a tuple throught the new `Identifier()` method.
For `ServicePrincipal`s (representing an OAuth2 client authenticated through `client_credentials` grant against Hydra), the subject will match the client-id of that specific OAuth2 client that generated the token. So in this case the identifier will be the value of the client-id coming form the `sub`claim.

For `UserPrincipal`s (representing a "human" user authenticated through the `authorization_code` grant against Hydra), the `email` claim will be used as the principal identifier.
This way we will be able to have OpenFGA tuples with a non-changing `user:` since both client-id and email uniquely identify an entity inside of the platform.

## Changes
- uniformed the use of the `isAdmin` context attribute (although right now is not used) and moved under authorization.go
- removed references to authorization header and dummy authentication in place before this
- now OpenFGA tuples are built using the `Identifier()` method on the `PrincipalInterface` to make authorization work on both CLI users and actual users
- auhtorization middlewares are moved after authentication middleware so authentication can come into play before and the authorizaion middlewares can rely on the authenticated principal object
- many tests are adjusted to suppor the changes in both authorization and authentication

# Breaking change
Anonymous authentication is not supported anymore. We can bring it back but we need to talk about what to put in and what not.

### DIFF
--- a/internal/authorization/middlewares.go
+++ b/internal/authorization/middlewares.go
@@ -1,11 +1,10 @@
 // Copyright 2024 Canonical Ltd.
-// SPDX-License-Identifier: AGPL
+// SPDX-License-Identifier: AGPL-3.0
 
 package authorization
 
 import (
 	"context"
-	"encoding/base64"
 	"encoding/json"
 	"fmt"
 	"net/http"
@@ -15,22 +14,12 @@ import (
 	"github.com/canonical/identity-platform-admin-ui/internal/http/types"
 	"github.com/canonical/identity-platform-admin-ui/internal/logging"
 	"github.com/canonical/identity-platform-admin-ui/internal/monitoring"
+	"github.com/canonical/identity-platform-admin-ui/pkg/authentication"
 )
 
 const (
-	// custom header for the time being
-	TOKEN_HEADER                 = "X-Authorization"
-	ADMIN_PRIVILEGE              = "privileged:superuser"
-	ADMIN_CTX       AdminContext = "adminCtx"
-	USER_CTX        UserContext  = "userCtx"
+	ADMIN_PRIVILEGE = "privileged:superuser"
 )
-
-type UserContext string
-type AdminContext string
-
-type User struct {
-	ID string
-}
 
 // Middleware is the monitoring middleware object implementing Prometheus monitoring
 type Middleware struct {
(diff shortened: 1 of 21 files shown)


shipperizer (approved):
code looks good, i'd say fixup the "adjust tests" commits into the respective prev commit

shipperizer (dismissed):

nsklikas (commented):
minor comments

internal/authorization/middlewares.go
```diff
@@ -49,38 +38,13 @@ type Middleware struct {
...
-func (mdw *Middleware) admin(r *http.Request) bool {
-	// TODO @shipperizer implement how to fetch user from cookie or header
-	user := mdw.transformToken(r.Header.Get(TOKEN_HEADER))
-
-	isAdmin, err := mdw.auth.Check(context.Background(), fmt.Sprintf("user:%s", user.ID), "admin", ADMIN_PRIVILEGE)
+func (mdw *Middleware) isAdmin(principal authentication.PrincipalInterface) bool {
+	isAdmin, err := mdw.auth.Check(context.Background(), fmt.Sprintf("user:%s", principal.Identifier()), "admin", ADMIN_PRIVILEGE)
 
 	return isAdmin && err == nil
 }
```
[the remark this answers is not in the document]
BarcoMasile: Honestly I don't know, I just renamed the function and used the `Identifier` method.
I would ask @shipperizer who definitely knows more on this than me 👼🏻

internal/authorization/middlewares.go
```diff
@@ -166,10 +130,11 @@ func (mdw *Middleware) Authorize() func(http.Handler) http.Handler {
 					return
 				}
 
-				user := mdw.user(r)
-				isAdmin := mdw.admin(r)
+				// if we got here then `principal` must be != nil
+				principal := authentication.PrincipalFromContext(r.Context())
```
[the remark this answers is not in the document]
BarcoMasile: I think it would make sense, even at the beginning of this.
Right now the panic would still happend due to invoking a method on a nil object, but an explicit panic would be way clearer. I'll add it, thanks!

internal/authorization/middlewares.go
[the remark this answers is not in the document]
shipperizer: if that s available in the authorizer interface then I'd say let's use it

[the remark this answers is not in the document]
BarcoMasile: It's in the AdminAuthorizer struct. I'll extract the interface and add it to the middleware object to use it
So we reuse existing function.
I'll push soon

internal/authorization/middlewares.go
```diff
@@ -132,7 +125,19 @@ func (mdw *Middleware) Authorize() func(http.Handler) http.Handler {
 
 				// if we got here then `principal` must be != nil
 				principal := authentication.PrincipalFromContext(r.Context())
-				isAdmin := mdw.isAdmin(principal)
+				if principal == nil {
+					// should never happen if authentication is configured correctly
+					mdw.logger.Fatalf("principal not available in context, cannot proceed with authorization")
```
shipperizer, line 130: we cannot panic when a request is in flight, we can return 500 though and log hard

internal/authorization/middlewares.go
```diff
@@ -132,7 +125,19 @@ func (mdw *Middleware) Authorize() func(http.Handler) http.Handler {
...
+					return
+				}
+
+				isAdmin, err := mdw.auth.Admin().CheckAdmin(r.Context(), principal.Identifier())
```
shipperizer, line 134: `isAdmin, err := mdw.auth.AdminAuthorizer.CheckAdmin(r.Context(), principal.Identifier())`

this would have worked the same i think, do we need it for testing purposes or we simply like interfaces?

internal/authorization/middlewares.go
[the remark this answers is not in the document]
shipperizer: we wouldnt recognize this as an error until someone logs in, which potentially might be after days

if we panic it should be at bootstrap

shipperizer (changes requested):

internal/authorization/middlewares.go
BarcoMasile: it's to fit the admin authorizer in the authorizer scenario without disrupting too much of the code

BarcoMasile: in order to work it needs to be exposed as a method on the interface, otherwise we would need to cast the interface to a concrete type

pkg/authentication/handlers.go
```diff
@@ -258,7 +258,11 @@ func (a *API) handleLogout(w http.ResponseWriter, r *http.Request) {
 }
 
 func (a *API) uiRedirect(w http.ResponseWriter, r *http.Request, nextTo string) {
-	redirect := ui.UIPrefix
+	redirect, err := url.JoinPath("/", a.contextPath, ui.UIPrefix)
+	if err != nil {
+		a.badRequest(w, err)
```
nsklikas, line 263: If we ever end up here, it means that the user has misconfigured the app. I think we should log it and return `500` instead

nsklikas (approved):
LGTM, only one pretty minor comment

shipperizer (approved):

pkg/authentication/handlers.go
BarcoMasile: Done! Thanks

shipperizer (approved):

nsklikas (approved):

PR #362, opened 2024-07-19
Use contextPath to redirect to UI

nsklikas:
The current logic is bugged, `r.URL.Path` does not always contain the full path (when running behind a reverse proxy, the proxy changes the path when forwarding the requests)

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -116,7 +116,8 @@ func serve() {
 	rulesConfig := rules.NewConfig(specs.RulesConfigMapName, specs.RulesConfigFileName, specs.RulesConfigMapNamespace, k8sCoreV1, externalConfig.OathkeeperPublic().ApiApi())
 
 	uiConfig := &ui.Config{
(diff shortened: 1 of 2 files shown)


pkg/authentication/handlers.go
```diff
@@ -258,7 +258,7 @@ func (a *API) handleLogout(w http.ResponseWriter, r *http.Request) {
 }
 
 func (a *API) uiRedirect(w http.ResponseWriter, r *http.Request, nextTo string) {
-	redirect := ui.UIPrefix
+	redirect := a.contextPath + ui.UIPrefix
```
shipperizer, line 261: we should make sure this is a valid uri, can we use http lib to parse and join?

nsklikas: Damn, that shouldn't be in here. Good catch

BarcoMasile (approved):

shipperizer (approved):

PR #368, opened 2024-07-31
WD-13638 - feat: log in via OIDC

huwshimi:
## Changes

- Set up component testing.
- Check if authenticated via the API.
- Login using OIDC.

https://warthogs.atlassian.net/browse/WD-13638

## QA

- Set up the UI and identity backend, following these instructions: https://github.com/canonical/identity-platform-admin-ui/wiki/Dev-UI-with-OIDC.
- Visit http://localhost:8000/api/v0/auth/logout to make sure you're logged out.
- Visit http://localhost:8000/ui
- You should see the login screen.
- Click the login button and after logging you in you should get returned to the UI which should show you as logged in.


ui/src/global.d.ts
```diff
@@ -0,0 +1 @@
+import "vitest-fetch-mock";
```
BarcoMasile, line 1: question: is this supposed to be under `src`?

huwshimi: Ah thanks this is actually no longer required as I fixed the typescript config in my [previous PR](https://github.com/canonical/identity-platform-admin-ui/commit/7781bcb13d2ea71c0b898c8ff47131334fd7df9e#diff-3ec3dd66f196bbc443564513e0afea83c0b280044551e1d29715f943a18147d3). Global definitions do need to be in the root, but don't get included in the build as they're just type annotations.

BarcoMasile (approved):

vladimir-cucu (approved):

PR #371, opened 2024-08-01
WD-13813 - refactor: migrate API calls to Axios

huwshimi:
## Changes

- Migrate API calls to Axios.
- Set API base path using Axios config.

This change means we can centrally handle 401 errors for all API calls and display the OIDC login if needed.

https://warthogs.atlassian.net/browse/WD-13813

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -53,6 +53,7 @@
     "@typescript-eslint/parser": "7.3.1",
     "@vitejs/plugin-react": "^4.2.1",
     "autoprefixer": "10.4.19",
+    "axios-mock-adapter": "2.0.0",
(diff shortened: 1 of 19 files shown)


BarcoMasile (dismissed):
LGTM: just one question, do you plan (in the future) to switch from promise to try/catch with async/await?

huwshimi:
> LGTM: just one question, do you plan (in the future) to switch from promise to try/catch with async/await?

I hadn't especially planned for that, but I don't think there's any reason why we couldn't.

vladimir-cucu (approved):

BarcoMasile (approved):

PR #375, opened 2024-08-12
WD-13639 - feat: log out with OIDC

huwshimi:
## Changes

- Add link to logout with OIDC.

## QA
- Log in.
- Click the Logout link at the bottom of the side nav.
- You should get redirected back from the logout URL to the UI and you should see the login button.

https://warthogs.atlassian.net/browse/WD-13639

### DIFF
--- a/ui/src/components/Navigation.tsx
+++ b/ui/src/components/Navigation.tsx
@@ -3,13 +3,14 @@ import { FC } from "react";
 import { NavLink, NavLinkProps } from "react-router-dom";
 import { Button } from "@canonical/react-components";
 import { GroupsLink, RolesLink } from "@canonical/rebac-admin";
+import { authURLs } from "api/auth";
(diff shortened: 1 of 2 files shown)


ui/src/components/Navigation.tsx
```diff
@@ -68,11 +69,10 @@ const Navigation: FC<Props> = ({ username, logout }) => {
                 nonInteractive: true,
               },
               <Button
+                element="a"
```
vladimir-cucu, line 72: Is there a specific reason why we make it an anchor element?

vladimir-cucu (approved):
Tested it locally and looks good!

ui/src/components/Navigation.tsx
huwshimi: Yep! We need to visit the logout URL for it to perform the logout.

shipperizer (approved):

PR #381, opened 2024-08-19
IAM-970 implement GroupService based on the Rebac lib

wood-push-melon:


pkg/groups/service.go
```diff
@@ -561,3 +565,359 @@ func NewService(ofga OpenFGAClientInterface, wpool pool.WorkerPoolInterface, tra
...
+}
+
+// PatchGroupIdentities assigns and removes the given resources.GroupIdentitiesPatchItem associated with the given group.
+func (s *V1Service) PatchGroupIdentities(ctx context.Context, groupId string, identityPatches []resources.GroupIdentitiesPatchItem) (bool, error) {
+	ctx, span := s.tracer.Start(ctx, "groups.V1Service.PatchGroupIdentities")
+	defer span.End()
+
+	if principal := authentication.PrincipalFromContext(ctx); principal == nil {
+		return false, v1.NewAuthorizationError("unauthorized")
+	}
```
shipperizer, line 723: we won't need this

pkg/groups/service.go
```diff
@@ -561,3 +565,359 @@ func NewService(ofga OpenFGAClientInterface, wpool pool.WorkerPoolInterface, tra
...
+
+// GetGroupRoles returns a page of resources.Role associated with the given group.
+func (s *V1Service) GetGroupRoles(ctx context.Context, groupId string, params *resources.GetGroupsItemRolesParams) (*resources.PaginatedResponse[resources.Role], error) {
+	ctx, span := s.tracer.Start(ctx, "groups.V1Service.GetGroupRoles")
+	defer span.End()
+
+	if principal := authentication.PrincipalFromContext(ctx); principal == nil {
+		return nil, v1.NewAuthorizationError("unauthorized")
+	}
+
```
shipperizer, line 770: same as comment above, middleware will take care of this

pkg/groups/service.go
```diff
@@ -561,3 +565,359 @@ func NewService(ofga OpenFGAClientInterface, wpool pool.WorkerPoolInterface, tra
...
+}
+
+// UpdateGroup updates the given resources.Group.
+//
+// Note: this is not implemented yet.
+func (s *V1Service) UpdateGroup(ctx context.Context, group *resources.Group) (*resources.Group, error) {
+	_, span := s.tracer.Start(ctx, "groups.V1Service.UpdateGroup")
+	defer span.End()
+
+	return nil, v1.NewNotImplementedError("service not implemented")
```
shipperizer, line 657: we have to implement this in v1 I believe

pkg/groups/service_test.go
```diff
@@ -1271,3 +1279,954 @@ func TestServiceRemovePermissions(t *testing.T) {
...
+	token := "eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJtb2NrLXN1YmplY3QiLCJhdWQiOiJtb2NrLWNsaWVudC1pZCIsIm5hbWUiOiJKb2huIERvZSIsImlhdCI6MTUxNjIzOTAyMn0.BdspASNsnxeXnqZXZnFnkvv-ClMq0U6X1gCIUrh9V7c"
+	principal, _ := authentication.NewJWKSTokenVerifier(mockProvider, "mock-client-id", mockTracer, mockLogger, mockMonitor).VerifyAccessToken(context.TODO(), token)
+
+	type testCase struct {
+		name           string
+		setupMocks     func()
+		contextSetup   func() context.Context
+		expectedResult []string
+		expectedError  error
+	}
```
shipperizer, line 1314: this looks like an interesting pattern, any chance you can explain a bit (as a pr comment) the benefits of the implementation?

wood-push-melon: yeah, basically it's like what the other tests do. Generally, a test could be divided into three steps: arrange, act, and assert. The `testCase` tries to extract the dynamic parts out of the fixed steps.

pkg/groups/service.go
```diff
@@ -561,3 +565,335 @@ func NewService(ofga OpenFGAClientInterface, wpool pool.WorkerPoolInterface, tra
...
+	}
+
+	r := &resources.PaginatedResponse[resources.Identity]{
+		Meta: resources.ResponseMeta{Size: len(identities)},
+		Data: make([]resources.Identity, 0, len(identities)),
+		Next: resources.Next{PageToken: &metaParam},
+	}
+
+	for _, identity := range identities {
+		r.Data = append(r.Data, resources.Identity{Id: &identity})
```
shipperizer, line 706: this might sound weird, but we might need to find a way to enhance these results with what we get from kratos, or at least find a way to add a link to the endpoint `getIdentity` that shows the right amount of data

i would argue that this for us is not a great endpoint due to the complexities around satisfying the aggregation of data (this would involve a call per identity object as i dont think kratos supports a "gimme all the identities in this filter")

pkg/groups/service.go
```diff
@@ -679,10 +679,6 @@ func (s *V1Service) GetGroupIdentities(ctx context.Context, groupId string, para
 	ctx, span := s.tracer.Start(ctx, "groups.V1Service.GetGroupIdentities")
 	defer span.End()
 
-	if principal := authentication.PrincipalFromContext(ctx); principal == nil {
-		return nil, v1.NewAuthorizationError("unauthorized")
-	}
-
```
BarcoMasile: Tagging @nsklikas since we added this pieces of code following one of his comments. This was done to be extra safe.
I am fine removing it, but I think we should agree since afaik there's diverging choices on this particular subject.

nsklikas (approved):

pkg/groups/service.go
wood-push-melon: @shipperizer Yeah, I remember we talked about this. I tried something similar to `GetIdentity` in identity services. Let me know if this is OK at this moment so I can go ahead and hit the unit test cases.

For the identity details handling, I guess we will come to a consistent way when we know what fields should be filled in the response. (although we never know the identity schema that the user might use.)

```golang
func (s *V1Service) GetGroupIdentities(ctx context.Context, groupId string, params *resources.GetGroupsItemIdentitiesParams) (*resources.PaginatedResponse[resources.Identity], error) {
	ctx, span := s.tracer.Start(ctx, "groups.V1Service.GetGroupIdentities")
	defer span.End()

	paginator := types.NewTokenPaginator(s.tracer, s.logger)
	if err := paginator.LoadFromString(ctx, *params.NextToken); err != nil {
		s.logger.Error(fmt.Sprintf("failed to parse the page token: %v", err))
	}

	identities, pageToken, err := s.core.ListIdentities(ctx, groupId, *params.NextToken)
	if err != nil {
		return nil, v1.NewUnknownError(fmt.Sprintf("failed to list identities for group %s: %v", groupId, err))
	}

	paginator.SetToken(ctx, GROUP_TOKEN_KEY, pageToken)
	metaParam, err := paginator.PaginationHeader(ctx)
	if err != nil {
		s.logger.Errorf("failed to create the pagination meta param: %v", err)
		metaParam = ""
	}

	r := &resources.PaginatedResponse[resources.Identity]{
		Meta: resources.ResponseMeta{Size: len(identities)},
		Data: make([]resources.Identity, 0, len(identities)),
		Next: resources.Next{PageToken: &metaParam},
	}

	for _, user := range identities {
		userParts := strings.SplitN(user, ":", 2)
		identityID := userParts[1]

		kratosIdentity, _, err := s.kratos.GetIdentityExecute(
			s.kratos.GetIdentity(ctx, identityID),
		)
		if err != nil {
			s.logger.Warnf("skipping identity %s due to error: %v", user, err)
			continue
		}

		traits, ok := kratosIdentity.Traits.(map[string]string)
		if !ok {
			traits = make(map[string]string)
		}

		identity := &resources.Identity{
			Id:    &kratosIdentity.Id,
			Email: traits["email"],
		}

		fullName := traits["name"]
		surnameIndex := strings.LastIndex(fullName, " ")

		if surnameIndex > 0 {
			firstName := strings.TrimSpace(fullName[:surnameIndex])
			lastName := strings.TrimSpace(fullName[surnameIndex:])
			identity.FirstName = &firstName
			identity.LastName = &lastName
		}

		r.Data = append(r.Data, *identity)
	}

	r.Meta.Size = len(r.Data)
	return r, nil
}
```

BarcoMasile (dismissed):

nsklikas (approved):

shipperizer (approved):

PR #382, opened 2024-08-20
IAM 968

shipperizer:
IAM-968: REBAC Identities service implementation

- **fix: update tracing signature**
- **fix: create openfga store to enhance basic client and offload core application logic**
- **feat: identities service implementation**

### DIFF
--- a/internal/openfga/interfaces.go
+++ b/internal/openfga/interfaces.go
@@ -9,7 +9,7 @@ import (
 	"github.com/openfga/go-sdk/client"
 )
 
-type OpenFGAClientInterface interface {
+type OpenFGACoreClientInterface interface {
 	GetAuthorizationModelId() (string, error)
 	CreateStore(context.Context) client.SdkClientCreateStoreRequestInterface
 	CreateStoreExecute(client.SdkClientCreateStoreRequestInterface) (*client.ClientCreateStoreResponse, error)
@@ -29,4 +29,13 @@ type OpenFGAClientInterface interface {
 	WriteExecute(client.SdkClientWriteRequestInterface) (*client.ClientWriteResponse, error)
 	ListObjects(context.Context) client.SdkClientListObjectsRequestInterface
 	ListObjectsExecute(client.SdkClientListObjectsRequestInterface) (*client.ClientListObjectsResponse, error)
+}
+
+// OpenFGAClientInterface is the interface used to decouple the OpenFGA store implementation
+type OpenFGAClientInterface interface {
+	ListObjects(context.Context, string, string, string) ([]string, error)
+	ReadTuples(context.Context, string, string, string, string) (*client.ClientReadResponse, error)
+	WriteTuples(context.Context, ...Tuple) error
+	DeleteTuples(context.Context, ...Tuple) error
+	Check(context.Context, string, string, string, ...Tuple) (bool, error)
 }
--- a/internal/openfga/stores.go
+++ b/internal/openfga/stores.go
@@ -1,0 +1,344 @@
+// Copyright 2024 Canonical Ltd.
+// SPDX-License-Identifier: AGPL-3.0
+
+package openfga
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"sync"
+
+	"github.com/canonical/identity-platform-admin-ui/internal/logging"
+	"github.com/canonical/identity-platform-admin-ui/internal/monitoring"
+	"github.com/canonical/identity-platform-admin-ui/internal/pool"
+
+	trace "go.opentelemetry.io/otel/trace"
+)
+
+const (
+	ASSIGNEE_RELATION = "assignee"
+	MEMBER_RELATION   = "member"
+	CAN_VIEW_RELATION = "can_view"
+)
+
+// TODO @shipperizer this is internal material, worth reusing it across the board
+// OpenFGAStore is an overarching store object to deal with OpenFGA entities, meant as a low level
+// object to perform cross cutting logic only relevant to the application, therefore doesn't deal with
+// user interpolations or returns fancy objects, that is offloaded to the service layer favouring reusability
+type OpenFGAStore struct {
+	ofga OpenFGAClientInterface
+
+	wpool pool.WorkerPoolInterface
+
+	tracer  trace.Tracer
+	monitor monitoring.MonitorInterface
+	logger  logging.LoggerInterface
+}
+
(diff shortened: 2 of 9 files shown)


pkg/identities/service.go
```diff
@@ -240,3 +249,514 @@ func NewService(kratos kClient.IdentityAPI, authz AuthorizerInterface, tracer tr
...
+	}
+
+	return true, nil
+}
+
+type Config struct {
+	Name         string
+	Namespace    string
+	K8s          coreV1.CoreV1Interface
+	OpenFGAStore OpenFGAStoreInterface
```
wood-push-melon, line 749: Just a **NONBLOCKING** question: why `Config` has a field `OpenFGAStore`? Seems that `Config` is closer to k8s configmap stuff.

pkg/identities/service.go
```diff
@@ -240,3 +249,514 @@ func NewService(kratos kClient.IdentityAPI, authz AuthorizerInterface, tracer tr
...
+		// needs to be validated as schemas might differ
+		body,
+	)
+
+	if err != nil {
+		return nil, v1.NewUnknownError(err.Error())
+	}
+
+	if len(ids.Identities) != 1 {
+		return nil, v1.NewInvalidRequestError("no identity created")
```
wood-push-melon, line 488: A very trivial thing: `no identity updated`.

And is it ok/better to check `len(ids.Identities) == 0`?

pkg/identities/service.go
```diff
@@ -240,3 +249,514 @@ func NewService(kratos kClient.IdentityAPI, authz AuthorizerInterface, tracer tr
...
+			i.LastName = &surname
+		}
+
+		r.Data = append(r.Data, i)
+	}
+
+	return r, nil
+}
+
+// // CreateIdentity creates a single Identity.
```
wood-push-melon, line 351: A trivial thing: double `//`

pkg/identities/service.go
shipperizer: it follows the same pattern as https://github.com/canonical/identity-platform-admin-ui/blob/main/pkg/schemas/service.go#L27-L32

ideally we never have more than 4/5 args in a constructor and use a config type

pkg/identities/service.go
shipperizer: issue is that the Service v0 returns always a list, in this case the result needs to be 1 or nothing

wood-push-melon (approved):

internal/openfga/interfaces.go
```diff
@@ -30,3 +30,12 @@ type OpenFGAClientInterface interface {
...
 	ListObjectsExecute(client.SdkClientListObjectsRequestInterface) (*client.ClientListObjectsResponse, error)
 }
+
+// OpenFGAClientInterface is the interface used to decouple the OpenFGA store implementation
+type OpenFGAClientInterface interface {
+	ListObjects(context.Context, string, string, string) ([]string, error)
+	ReadTuples(context.Context, string, string, string, string) (*client.ClientReadResponse, error)
+	WriteTuples(context.Context, ...Tuple) error
+	DeleteTuples(context.Context, ...Tuple) error
+	Check(context.Context, string, string, string, ...Tuple) (bool, error)
```
BarcoMasile, line 40: question: is this intended as a "Find" store operation? Or should we put it under the Core interface since it's a check on the tuples?

I'm good either way

internal/openfga/stores.go
```diff
@@ -0,0 +1,343 @@
...
+	return nil
+}
+
+// AssignGroups assigns groups to an "assignee"able resource (user, group#member)
+func (s *OpenFGAStore) AssignGroups(ctx context.Context, assigneeID string, groupIDs ...string) error {
+	ctx, span := s.tracer.Start(ctx, "openfga.OpenFGAStore.AssignGroups")
+	defer span.End()
+
+	// preemptive check to verify if all Groups to be assigned are accessible by the user
+	// needs to happen separately
```
BarcoMasile, line 137: question: this seems like a `// TODO`, should have an issue to track this or do you plan on adding this check in this PR?

internal/openfga/stores.go
```diff
@@ -0,0 +1,343 @@
...
+	monitor monitoring.MonitorInterface
+	logger  logging.LoggerInterface
+}
+
+// ListViewableRoles returns all the roles a specific "assignee"able resource (user, group#member, role#assignee) is linked to (using "can_view" OpenFGA relation)
+func (s *OpenFGAStore) ListViewableRoles(ctx context.Context, ID string) ([]string, error) {
+	ctx, span := s.tracer.Start(ctx, "openfga.OpenFGAStore.ListViewableRoles")
+	defer span.End()
+
+	roles, err := s.ofga.ListObjects(ctx, ID, "can_view", "role")
```
BarcoMasile, line 43: `"can_view"` is the only relation that doesn't have a constant

BarcoMasile (approved):

internal/openfga/interfaces.go
shipperizer: interface is only meant to be used by https://github.com/canonical/identity-platform-admin-ui/pull/382/files#diff-0c2915818e34863217890b989cf6bcaf87f9daaab08e908a2825257a3aab38e8R29

and that is basically our client wrapper around the `core` openfga client https://github.com/canonical/identity-platform-admin-ui/pull/382/files#diff-5aa2f2c8ad62191825fc8a94d7a9e265076d95f62428b8e41f6012e6974035c1R12

now naming convention is terrible here, as if someone decides to use this interface it will get confused


we cannot put those 2 checks in the same interface as this is the "user friendly" one, the `find` method should be implemented on the store but it was not needed atm

plan is to add it when we refactor roles and groups

internal/openfga/stores.go
shipperizer: https://github.com/canonical/identity-platform-admin-ui/issues/387

BarcoMasile (approved):

nsklikas (approved):
lgtm overall, just one minor suggestion. Feel free to merge without changing it

internal/openfga/stores.go
```diff
@@ -0,0 +1,344 @@
...
+		return nil, "", err
+	}
+
+	permissions := make([]Permission, 0)
+
+	for _, t := range r.GetTuples() {
+		// if relation doesn't start with can_ it means it's not a permission (see #assignee)
+		if !strings.HasPrefix(t.Key.Relation, "can_") {
+			continue
+		}
```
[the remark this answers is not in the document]
shipperizer: fair game, let me fix it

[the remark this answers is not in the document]
nsklikas: sure

PR #396, opened 2024-09-02
Add capabilities service

nsklikas:
IAM-971

### DIFF
--- a/pkg/capabilities/service.go
+++ b/pkg/capabilities/service.go
@@ -1,0 +1,44 @@
+// Copyright 2024 Canonical Ltd.
+// SPDX-License-Identifier: AGPL-3.0
+
+package capabilities
+
+import "github.com/canonical/rebac-admin-ui-handlers/v1/resources"
+
+type V1Service struct{}
+
+const APIPrefix = "/api/v1"
+
+func (s *V1Service) ListCapabilities() ([]resources.Capability, error) {
+	result := []resources.Capability{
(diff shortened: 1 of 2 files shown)


pkg/capabilities/service.go
```diff
@@ -0,0 +1,44 @@
...
+		{Endpoint: APIPrefix + "/roles", Methods: []resources.CapabilityMethods{"GET", "POST"}},
+		{Endpoint: APIPrefix + "/roles/{id}", Methods: []resources.CapabilityMethods{"GET", "PUT", "DELETE"}},
+		{Endpoint: APIPrefix + "/roles/{id}/entitlements", Methods: []resources.CapabilityMethods{"GET", "PATCH"}},
+		{Endpoint: APIPrefix + "/entitlements", Methods: []resources.CapabilityMethods{"GET"}},
+		{Endpoint: APIPrefix + "/entitlements/raw", Methods: []resources.CapabilityMethods{"GET"}},
+		{Endpoint: APIPrefix + "/resources", Methods: []resources.CapabilityMethods{"GET"}},
+	}
+
+	return result, nil
+}
```
shipperizer, line 38: instead of this i think we go with the `inferCapabilities` that is the default on on the handler if the Capabilities service is nil

we need to open a bug in there tp make sure that supports the prefix

nsklikas:
Opened issue https://github.com/canonical/rebac-admin-ui-handlers/issues/25

nsklikas:
Closing this. The library should append the prefix baseURL to the auto-generated capabilities as indicated in the issue opened.

PR #402, opened 2024-09-06
feat: add entitlements service by Rebac

wood-push-melon:


pkg/entitlements/service.go
```diff
@@ -0,0 +1,145 @@
...
+		return nil, v1.NewAuthorizationError("unauthorized")
+	}
+
+	results := make(chan *pool.Result[any], len(ObjectTypes))
+	wg := sync.WaitGroup{}
+	wg.Add(len(ObjectTypes))
+
+	for _, objectType := range ObjectTypes {
+		_, err := s.wpool.Submit(
+			s.listEntitlementsByObjectTypeFunc(ctx, fmt.Sprintf("user:%s", principal.Identifier()), objectType, *params.Filter),
```
shipperizer, line 51: in this case instead of returning a proper tuple we should parse the schema and have 

`client`, `can_view`, `user,group#member,role#assignee`
`client`, `can_create`, `user,group#member,role#assignee`
`client`, `can_edit`, `user,group#member,role#assignee`

so no real tuples but the definition of the types

wood-push-melon: Completely misunderstood what this API tries to do 😅. I will change it.

pkg/entitlements/service.go
```diff
@@ -0,0 +1,112 @@
...
+	tracer  trace.Tracer
+	monitor monitoring.MonitorInterface
+	logger  logging.LoggerInterface
+}
+
+func (s *V1Service) ListEntitlements(ctx context.Context, params *resources.GetEntitlementsParams) ([]resources.EntitlementSchema, error) {
+	ctx, span := s.tracer.Start(ctx, "entitlements.V1Service.ListEntitlements")
+	defer span.End()
+
+	authModel, err := s.ofga.ReadModel(ctx)
```
shipperizer, line 29: model is immutable so probably good to make this an attribute if the struct

pkg/entitlements/service.go
```diff
@@ -0,0 +1,112 @@
...
+			receivers := buildReceivers(relationReferences)
+			entitlementSchema := resources.EntitlementSchema{
+				Entitlement:  relation,
+				EntityType:   typeDef.Type,
+				ReceiverType: receivers,
+			}
+
+			entitlementSchemas = append(entitlementSchemas, entitlementSchema)
+		}
+	}
```
shipperizer, line 57: following up on point above, we should probably cache this so we don't repeat it every time it gets called

shipperizer (commented):
model is immutable, so I'd say we make this an attribute of the struct

shipperizer (approved):

BarcoMasile (approved):

PR #412, opened 2024-09-13
IAM-1054 add the create-identity CLI

wood-push-melon:
The pull request tries to add the CLI `create-identity` which aligns with what the corresponding HTTP API does.

### DIFF
--- a/cmd/createIdentity.go
+++ b/cmd/createIdentity.go
@@ -1,0 +1,146 @@
+package cmd
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"os"
+
+	"github.com/canonical/identity-platform-admin-ui/internal/authorization"
+	"github.com/canonical/identity-platform-admin-ui/internal/config"
+	"github.com/canonical/identity-platform-admin-ui/internal/kratos"
+	"github.com/canonical/identity-platform-admin-ui/internal/logging"
+	"github.com/canonical/identity-platform-admin-ui/internal/mail"
+	"github.com/canonical/identity-platform-admin-ui/internal/monitoring"
+	"github.com/canonical/identity-platform-admin-ui/internal/openfga"
+	"github.com/canonical/identity-platform-admin-ui/internal/pool"
+	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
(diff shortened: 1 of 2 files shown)


cmd/createIdentity.go
```diff
@@ -0,0 +1,146 @@
...
+	"github.com/canonical/identity-platform-admin-ui/internal/pool"
+	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
+	"github.com/canonical/identity-platform-admin-ui/pkg/identities"
+	"github.com/kelseyhightower/envconfig"
+	kClient "github.com/ory/kratos-client-go"
+	"github.com/spf13/cobra"
+)
+
+var createIdentityCmd = &cobra.Command{
+	Use:   "create-identity [jsonfile]",
```
BarcoMasile, line 26: super small suggestion, feel free to ignore:
```suggestion
	Use:   "create-identity [path/to/file.json]",
```

so we explicit it's a path

BarcoMasile (approved):
Well done!

nsklikas (dismissed):
LGTM. It feels a little awkward to have to define these many env vars to call a CLI command, but there is not much we can do.

cmd/createIdentity.go
```diff
@@ -0,0 +1,146 @@
...
+	},
+	"credentials"": {
+		"password": {
+			"config": {
+				"password": "bar"
+			}
+		}
+	}
+}
+EOF
```
nsklikas, line 44: thought: Would it make sense to simplify this by asking the user to inject it via command args? eg:
`admin-ui create user --password 123 --schema-id default --trait email=foo@example.com --trait name=Bob --trait phone=1234`

wood-push-melon: yeah, however it would be a little difficult to represent nested traits I guess (correct me if I'm wrong). And the `trait` argument might go with a very long list.

We can update it whenever we find json input does not work conveniently. (changes will be limited to the argument parsing parting.)

cmd/createIdentity.go
wood-push-melon: let me update it.

nsklikas (approved):

shipperizer (approved):

PR #419, opened 2024-09-19
feat: introduce hierarchy for can_relations

BarcoMasile:
## Description
This PR introduces a hierarchy for all `can_*` relations in the OpenFGA authorization model.
The hierarchy is as follows
- if a user can delete an object (top permission for can_* permission) then the user can also edit it
- if a user can edit an object, then the user can also view it
- if a user can view an object, the user doesn't inherit additional permissions from this relation only

So: `can_delete` >> `can_edit` >> `can_view`
Can create is not touched by this since it gets special treatment in the implementation.

## Why
This is done with the intention of simplifying both user experience and developer life when it comes to reasoning about user authorization based on entitlements.
This way we can rely on some simple implications to treat entitlements. Also, this allows us to not crowd the OpenFGA store with "same user" tuples.

### DIFF
--- a/internal/authorization/schema.openfga
+++ b/internal/authorization/schema.openfga
@@ -14,9 +14,9 @@ type role
 
     define can_create: [user, role#assignee, group#member] or admin from privileged
     define can_delete: [user, role#assignee, group#member] or admin from privileged
-    define can_edit: [user, role#assignee, group#member] or admin from privileged
-    define can_view: [user, user:*, role#assignee, group#member] or admin from privileged
-    
+    define can_edit: [user, role#assignee, group#member] or can_delete or admin from privileged
+    define can_view: [user, user:*, role#assignee, group#member] or can_edit or admin from privileged
+
 type group
  relations
     define privileged: [privileged]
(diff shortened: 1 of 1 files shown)


nsklikas (dismissed):
No problem with this, but I have a question:
What is the upgrade path for this? This is a question for the charm, but I am curious to know

BarcoMasile:
> No problem with this, but I have a question: What is the upgrade path for this? This is a question for the charm, but I am curious to know

@nsklikas it should suffice to just apply the new authorization model. It should not cause conflicts (OpenFGA handles multiple authorization model weird for what I remember).
We could technically dump the database tuples content, and re-apply the content after the upgrade, since the previous tuples would be 100% valid even after the change.
But this should be investigated, thanks for raising this!

shipperizer:
> No problem with this, but I have a question: What is the upgrade path for this? This is a question for the charm, but I am curious to know

afaik the upgrade of the model brings the tuples with it, we would have to trigger the right relationship refresh to communicate the app(s) that the model id has changed

in our specific case, this would trigger a clash as we do the model validation at runtime, so it's not something that can be orchestrated without scheduled downtime

question is: do we validate the model consistency or we rely on openfga blindly?

internal/authorization/schema.openfga
```diff
@@ -14,71 +14,71 @@ type role
 
     define can_create: [user, role#assignee, group#member] or admin from privileged
     define can_delete: [user, role#assignee, group#member] or admin from privileged
-    define can_edit: [user, role#assignee, group#member] or admin from privileged
-    define can_view: [user, user:*, role#assignee, group#member] or admin from privileged
-    
+    define can_edit: [user, role#assignee, group#member] or admin from privileged or can_delete
```
shipperizer, line 17:
```suggestion
    define can_edit: [user, role#assignee, group#member] or can_delete or admin from privileged 
```


dont hate me

BarcoMasile: I do but I'll change it anyway 😃

shipperizer:
side note, when we create a resource what tuples do we assign ?is it can view and can edit?if yes we can drop the first

BarcoMasile:
> side note, when we create a resource what tuples do we assign ?is it can view and can edit?if yes we can drop the first
> 
> https://github.com/canonical/identity-platform-admin-ui/blob/main/internal%2Fauthorization%2Fresource_manager.go#L23

@shipperizer for groups and roles we only add `can_view` for the newly created object

shipperizer:
> > side note, when we create a resource what tuples do we assign ?is it can view and can edit?if yes we can drop the first
> > 
> > https://github.com/canonical/identity-platform-admin-ui/blob/main/internal%2Fauthorization%2Fresource_manager.go#L23
> 
> @shipperizer for groups and roles we only add `can_view` for the newly created object

was more concerned about the other non OpenFGA backed resources, but, for roles and groups are we ok with can view only?

PR #420, opened 2024-09-20
WD-15023 - chore: set up MJML

huwshimi:
## Changes

Set up MJML email template builder (note: no templates have been added yet).

https://warthogs.atlassian.net/browse/WD-15023

### DIFF
--- a/internal/mail/Makefile
+++ b/internal/mail/Makefile
@@ -1,8 +1,0 @@
-MJML=npx mjml
-MJML_FILES=$(wildcard mjml/*.mjml)
-HTML_DIR=html
-
-build:
-	$(foreach file, $(MJML_FILES), $(MJML) $(file) --config.minify --config.useMjmlConfigOptions -o $(HTML_DIR)/$(basename $(notdir $(file))).html;)
-
-.PHONY=build
--- a/internal/mail/Makefile
+++ b/internal/mail/Makefile
@@ -1,0 +1,8 @@
+MJML=npx mjml
+MJML_FILES=$(wildcard mjml/*.mjml)
+HTML_DIR=html
+
+build:
+	$(foreach file, $(MJML_FILES), $(MJML) $(file) --config.minify --config.useMjmlConfigOptions -o $(HTML_DIR)/$(basename $(notdir $(file))).html;)
+
+.PHONY=build
--- a/internal/mail/.mjmlconfig
+++ b/internal/mail/.mjmlconfig
@@ -1,8 +1,0 @@
-{
-  "options": {
-    "minifyOptions": {
-      "minifyCSS": true
-    }
-  },
-  "packages": []
-}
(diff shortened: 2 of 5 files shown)


internal/mail/Makefile
huwshimi, line 1: Is it OK for these files to live in this directory?

shipperizer (commented):
I was left with the impression we were using backend templates with `.gotpl`

what did I miss?how are we going to operate with the templates from a backend perspective, or is the frontend doing the hard work and hitting the email endpoint with all the info it needs (risky if we pass the authentication code or password)?

cc @BarcoMasile

huwshimi:
> I was left with the impression we were using backend templates with `.gotpl`
> 
> what did I miss?how are we going to operate with the templates from a backend perspective, or is the frontend doing the hard work and hitting the email endpoint with all the info it needs (risky if we pass the authentication code or password)?
> 

The backend loads the templates e.g. https://github.com/canonical/identity-platform-admin-ui/blob/main/internal/mail/html/user-invite.html. However, we need a way to generate those templates, which we'll do using MJML (HTML emails use an HTML style from circa 2000 e.g. inline styles etc. and MJML streamlines generating the code and making it compatible with the various mail clients etc.)

shipperizer:
> > I was left with the impression we were using backend templates with `.gotpl`
> > what did I miss?how are we going to operate with the templates from a backend perspective, or is the frontend doing the hard work and hitting the email endpoint with all the info it needs (risky if we pass the authentication code or password)?
> 
> The backend loads the templates e.g. https://github.com/canonical/identity-platform-admin-ui/blob/main/internal/mail/html/user-invite.html. However, we need a way to generate those templates, which we'll do using MJML (HTML emails use an HTML style from circa 2000 e.g. inline styles etc. and MJML streamlines generating the code and making it compatible with the various mail clients etc.)

i m confused, need to see the whole thing working

BarcoMasile:
> > I was left with the impression we were using backend templates with `.gotpl`
> > what did I miss?how are we going to operate with the templates from a backend perspective, or is the frontend doing the hard work and hitting the email endpoint with all the info it needs (risky if we pass the authentication code or password)?
> 
> The backend loads the templates e.g. https://github.com/canonical/identity-platform-admin-ui/blob/main/internal/mail/html/user-invite.html. However, we need a way to generate those templates, which we'll do using MJML (HTML emails use an HTML style from circa 2000 e.g. inline styles etc. and MJML streamlines generating the code and making it compatible with the various mail clients etc.)

@huwshimi @shipperizer 
To clarify: yes we use backend templates from the go standard library.
If @huwshimi prefers to do it with MJML it's fine, but I'd ask you to please put this somewhere else in the UI folder (and handle that subfolder as you wish), and refactor the Makefile to make sure that only the minified html (generated via this tool) ends up in the templates folder. This way only the minified html produced by MJML ends up in the backend build. Also, make sure the 3 go template variables are present after MJML generates the html.
Is that ok for you?

huwshimi:
@BarcoMasile I've moved these files to the UI dir.

BarcoMasile (approved):
Thank you @huwshimi

shipperizer (approved):

PR #426, opened 2024-09-25
IAM-951 fix broken href for non-empty context-path

### LINKED ISSUE #350: React App routes and hrefs are broken
## Description
Right now the Frontend app gets served by the UI correctly at `${contextPath}/ui` (UI_PREFIX from now on).
But right after it gets loaded things start to break.
Assets URLs are not correct, where they don't present the expected UI_PREFIX but they end up starting from the root `/`.

Perform the login at `/api/v0/auth`, wait for the redirection after login:
image

As you can see the assets are not downloaded correctly. Upon further inspection, you can see that the URL requested for each asset doens't include the `/ui` prefix.

Go into the browser and add a trailing slash to the URL and see the following (Success):
image
image

**EDIT** after chat with @nsklikas :
This is fixed correctly by changing the makefile under the ui folder to present
`$(YARN) build --base=/ui`
Unfortunately this will break the Charm since it's deployed behind Traefik.

BarcoMasile:
## Description
Switch to html/template for rendering context path dynamically for `index.html`
This way when serving Admin UI under a context path we don't break the UI.
UI still expects to be under `/context-path(if present)/ui`.

## N.B.
We still need web team intervention to make the produced `index.html` to have the following prefix for its assets: `{{ . }}ui`

Example
```html
<head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1" />

    <title>Identity platform</title>
    <link rel="shortcut icon" href="https://assets.ubuntu.com/v1/49a1a858-favicon-32x32.png" type="image/x-icon" />

    <script>const global = globalThis;</script>
    <script type="module" crossorigin src="{{ . }}ui/assets/index-D2_UOyoG.js"></script>
    <link rel="stylesheet" crossorigin href="{{ . }}ui/assets/index-tyeITY-6.css">
  </head>
```

This PR was manually tested simulating a fake context-path under which the UI could be served. See the example with `with-context-path` used as a fake path. The index html file GETs the right path for the assets.

pr-context-path

### Fixes
Fixes #350

### DIFF
--- a/pkg/ui/handlers.go
+++ b/pkg/ui/handlers.go
@@ -5,6 +5,7 @@ package ui
 
 import (
 	"encoding/json"
+	"html/template"
 	"io/fs"
 	"net/http"
 	"net/url"
@@ -18,7 +19,10 @@ import (
 	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
 )
 
-const UIPrefix = "/ui"
+const (
+	UIPrefix      = "/ui"
+	indexTemplate = "index.html"
+)
 
 type Config struct {
 	DistFS      fs.FS
@@ -26,8 +30,9 @@ type Config struct {
 }
 
 type API struct {
-	fileServer  http.Handler
 	contextPath string
+	fileServer  http.Handler
+	distFS      fs.FS
 
 	tracer  tracing.TracingInterface
 	monitor monitoring.MonitorInterface
@@ -39,19 +44,23 @@ func (a *API) RegisterEndpoints(mux *chi.Mux) {
 		path, err := url.JoinPath("/", a.contextPath, UIPrefix, "/")
 		if err != nil {
 			a.logger.Error("Failed to construct path: ", err)
-			w.WriteHeader(http.StatusInternalServerError)
-			json.NewEncoder(w).Encode(
-				types.Response{
-					Status:  http.StatusInternalServerError,
-					Message: err.Error(),
-				},
-			)
+			a.internalServerErrorResponse(w, err)
 			return
 		}
 		http.Redirect(w, r, path, http.StatusMovedPermanently)
 	})
 	mux.Get(UIPrefix+"/", a.uiFiles)
 	mux.Get(UIPrefix+"/*", a.uiFiles)
+}
+
+func (a *API) internalServerErrorResponse(w http.ResponseWriter, err error) {
+	w.WriteHeader(http.StatusInternalServerError)
+	json.NewEncoder(w).Encode(
+		types.Response{
+			Status:  http.StatusInternalServerError,
+			Message: err.Error(),
+		},
+	)
 }
 
 func (a *API) uiFiles(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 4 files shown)


nsklikas (commented):
I don't really love this, but if it works, it works.

I have one question though:
If we are going to start templating/injecting the returned html, why don't we set the [html base tag](https://www.w3schools.com/tags/tag_base.asp) instead of editing paths? This seems much safer and extendable. Something like this:
```
diff --git a/pkg/ui/handlers.go b/pkg/ui/handlers.go
index b9b0a3e..842c9d8 100644
--- a/pkg/ui/handlers.go
+++ b/pkg/ui/handlers.go
@@ -102,10 +102,7 @@ func (a *API) uiFiles(w http.ResponseWriter, r *http.Request) {
                w.Header().Set("Cache-Control", "no-cache, no-store, must-revalidate")
                w.WriteHeader(http.StatusOK)
 
-               normContextPath := a.contextPath
-               if !strings.HasSuffix(normContextPath, "/") {
-                       normContextPath += "/"
-               }
+               normContextPath := r.URL.JoinPath(a.contextPath, UIPrefix, "/")
 
                err = t.Execute(w, normContextPath)
                if err != nil {
diff --git a/ui/index.html b/ui/index.html
index 4dd324f..ea803e4 100644
--- a/ui/index.html
+++ b/ui/index.html
@@ -4,6 +4,7 @@
   <head>
     <meta charset="UTF-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <base href="{{ . }}" target="_blank">
 
     <title>Identity platform</title>
     <link rel="shortcut icon" href="https://assets.ubuntu.com/v1/49a1a858-favicon-32x32.png" type="image/x-icon" />
```

BarcoMasile:
> I don't really love this, but if it works, it works.
> 
> I have one question though: If we are going to start templating/injecting the returned html, why don't we set the [html base tag](https://www.w3schools.com/tags/tag_base.asp) instead of editing paths? This seems much safer and extendable.

I think it's a good idea honestly, but from my notes from the meeting with @huwshimi we explicitly chose not to use the base path (I may have misinterpreted in this case though).

If it works, I am open to it and actually think it's better (as you said).
I would leave the decision to @huwshimi since his team owns the frontend part and including also the `/ui` part in the context will have some changes over there too.

huwshimi:
> > I don't really love this, but if it works, it works.
> > I have one question though: If we are going to start templating/injecting the returned html, why don't we set the [html base tag](https://www.w3schools.com/tags/tag_base.asp) instead of editing paths? This seems much safer and extendable.
> 
> I think it's a good idea honestly, but from my notes from the meeting with @huwshimi we explicitly chose not to use the base path (I may have misinterpreted in this case though).

I think the previous issue with base was that it didn't help us resolve the issue with a `/prefix/ui` path with no trailing slash. Now that we have the template var this seems like the cleanest approach.

nsklikas:
The main problem was that url's with more path parts were not allowed. E.g. `/something/ui/clients/client_info` would not work, because the relative path hack would try to fetch the assets from `/something/ui/clients/assets/...` instead of `/something/ui/assets/...`

BarcoMasile:
@huwshimi @nsklikas 
Nice guys.
So are we good to proceed with merging?

huwshimi:
> @huwshimi @nsklikas Nice guys. So are we good to proceed with merging?

Fine from my side.

BarcoMasile:
> > @huwshimi @nsklikas Nice guys. So are we good to proceed with merging?
> 
> Fine from my side.

@huwshimi nice, can you approve?
I'll wait for @nsklikas feedback

BarcoMasile:
Just rebased on main.

huwshimi:
> Just rebased on main.

It looks like this removed my changes. Would you like me to push them again?

BarcoMasile:
> > Just rebased on main.
> 
> It looks like this removed my changes. Would you like me to push them again?
@huwshimi 
Yes please. Sorry, I don't really know how that happened

huwshimi:
> > > Just rebased on main.
> > 
> > 
> > It looks like this removed my changes. Would you like me to push them again?
> 
> @huwshimi Yes please. Sorry, I don't really know how that happened

Done.

nsklikas:
Just to be clear, @huwshimi I am curious why you think that the current approach is cleaner than the one proposed at https://github.com/canonical/identity-platform-admin-ui/pull/426#pullrequestreview-2332914123

nsklikas (approved):

huwshimi (approved):

huwshimi:
> Just to be clear, @huwshimi I am curious why you think that the current approach is cleaner than the one proposed at [#426 (review)](https://github.com/canonical/identity-platform-admin-ui/pull/426#pullrequestreview-2332914123)

(as discussed elsewhere) this is using `<base>` as suggested. I'll also investigate dropping the `basePath` from the UI client after this lands.

PR #427, opened 2024-09-25
feat: add the send invitation email cli command for admin to compensate user invitation email failure

wood-push-melon:
This pull request tries to add a CLI command for the admins to compensate for email sending failure in the user invitation flow.

### DIFF
--- a/cmd/createEmail.go
+++ b/cmd/createEmail.go
@@ -1,0 +1,119 @@
+package cmd
+
+import (
+	"context"
+	"fmt"
+	"os"
+
+	"github.com/canonical/identity-platform-admin-ui/internal/authorization"
+	"github.com/canonical/identity-platform-admin-ui/internal/config"
+	"github.com/canonical/identity-platform-admin-ui/internal/kratos"
+	"github.com/canonical/identity-platform-admin-ui/internal/logging"
+	"github.com/canonical/identity-platform-admin-ui/internal/mail"
+	"github.com/canonical/identity-platform-admin-ui/internal/monitoring"
+	"github.com/canonical/identity-platform-admin-ui/internal/openfga"
+	"github.com/canonical/identity-platform-admin-ui/internal/pool"
+	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
+	"github.com/canonical/identity-platform-admin-ui/pkg/identities"
+	"github.com/kelseyhightower/envconfig"
+	"github.com/spf13/cobra"
+)
+
+var createEmailCmd = &cobra.Command{
+	Use:   "create-email [identity-id]",
+	Short: "Send an invitation email for an identity",
+	Example: `Send an invitation email:
+
+identity-platform-admin-ui create-email d0bebe60-f4fa-4bdc-b6d6-dff9eb42d287`,
+	Long: `Send an invitation email for an identity if previous email sending efforts failed.`,
+	Args: cobra.ExactArgs(1),
+	Run:  createEmailHandler,
+}
+
+func init() {
+	rootCmd.AddCommand(createEmailCmd)
+}
+
+func createEmailHandler(cmd *cobra.Command, args []string) {
+	ctx := context.Background()
+
+	logger := logging.NewNoopLogger()
+	tracer := tracing.NewNoopTracer()
+	monitor := monitoring.NewNoopMonitor("", logger)
+	wpool := pool.NewWorkerPool(1, tracer, monitor, logger)
+	defer wpool.Stop()
+
+	// Initialize environment variables
+	specs, err := initializeEnv()
+	if err != nil {
+		panic(err)
+	}
+
+	// Initialize identity service
+	service := initializeIdentityService(specs, logger, tracer, monitor, wpool)
+
+	// Get the identity
+	identityID := args[0]
+	identityData, err := service.GetIdentity(ctx, identityID)
+	if err != nil {
+		panic(fmt.Errorf("failed to get the identity: %v", err))
+	}
+
+	ids := identityData.Identities
+	if len(ids) == 0 {
+		cmd.PrintErrln("No identity found for identity:", identityID)
+		os.Exit(1)
+	}
+
+	identity := &ids[0]
+
+	// Send identity creation email
+	err = service.SendUserCreationEmail(ctx, identity)
+	if err != nil {
+		panic(fmt.Errorf("failed to send the identity creation email: %v", err))
+	}
+
+	cmd.Printf("Invitation email sent for identity: %s\n", identity.GetId())
+}
+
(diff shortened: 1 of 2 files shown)


cmd/createEmail.go
```diff
@@ -0,0 +1,119 @@
...
+	// Send identity creation email
+	err = service.SendUserCreationEmail(ctx, identity)
+	if err != nil {
+		panic(fmt.Errorf("failed to send the identity creation email: %v", err))
+	}
+
+	cmd.Printf("Invitation email sent for identity: %s\n", identity.GetId())
+}
+
+func initializeEnv() (*config.EnvSpec, error) {
```
wood-push-melon, line 79: Not very sure where is a better place to put `initializeEnv` and `initializeIdentityService`.

Let me know if there is a better place.

cmd/createEmail.go
```diff
@@ -0,0 +1,119 @@
...
+	"github.com/canonical/identity-platform-admin-ui/internal/openfga"
+	"github.com/canonical/identity-platform-admin-ui/internal/pool"
+	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
+	"github.com/canonical/identity-platform-admin-ui/pkg/identities"
+	"github.com/kelseyhightower/envconfig"
+	"github.com/spf13/cobra"
+)
+
+var createEmailCmd = &cobra.Command{
+	Use:   "create-email [identity-id]",
```
wood-push-melon, line 23: `create-email` cmd name is a bit odd. The reason for doing this is for CLI future improvements (these all fall into the `create` subcommand)

```
identity-platform-admin-ui create fga-model
identity-platform-admin-ui create admin
identity-platform-admin-ui create identity
identity-platform-admin-ui create email
```

Also if there are new types of emails coming up in the future, we could add cmd flags to include the cases.

cmd/createEmail.go
```diff
@@ -0,0 +1,119 @@
...
+	ids := identityData.Identities
+	if len(ids) == 0 {
+		cmd.PrintErrln("No identity found for identity:", identityID)
+		os.Exit(1)
+	}
+
+	identity := &ids[0]
+
+	// Send identity creation email
+	err = service.SendUserCreationEmail(ctx, identity)
```
wood-push-melon, line 71: Only checked the condition when identity is not found.

There might be a (edge) case that:
1. the first email-sending attempt in the user creation flow times out, but actually the email is sent successfully at the end (due to the nature of networking in the distributed system)
2. the admin run this cli to send the email again.
3. the user email box receives two emails without guaranteed order ("thanks" again to the networking)

I am not sure if the recovery code and invitation link need to be invalidated in the first email (from what I understand from the code, it does not need to be. But let me know if I'm wrong).

cmd/createEmail.go
```diff
@@ -0,0 +1,119 @@
...
+	}
+
+	// Initialize identity service
+	service := initializeIdentityService(specs, logger, tracer, monitor, wpool)
+
+	// Get the identity
+	identityID := args[0]
+	identityData, err := service.GetIdentity(ctx, identityID)
+	if err != nil {
+		panic(fmt.Errorf("failed to get the identity: %v", err))
```
pik4ez-canonical, line 59: What is the reason to `panic` in cases like this?

pik4ez-canonical (approved):
LGTM.

I would assume that `panic` shouldn't be used wherever possible to gracefully handle the error (read as: most of the cases). But I see that currently it is widely used in the Admin UI repo, so shouldn't be a blocker for this PR. Most likely it's something that we may consider as a tech debt.

cmd/createEmail.go
```diff
@@ -0,0 +1,119 @@
...
+	// Set up mail service
+	mailConfig := mail.NewConfig(
+		specs.MailHost,
+		specs.MailPort,
+		specs.MailUsername,
+		specs.MailPassword,
+		specs.MailFromAddress,
+		specs.MailSendTimeoutSeconds,
+	)
+	mailService := mail.NewEmailService(mailConfig, tracer, monitor, logger)
```
nsklikas, line 116: I don't like it that you have to specify all these env vars to run a CLI command, when some of them may not be needed (e.g. I don't think you need to connect to openfga to send an email). But with our current implementation this will only be called through juju/pebble, which means that the environment will already be there.

If you wanted to use this as a CLI from your local machine, then this would prove a little annoying or even unusable. But I guess we can refactor this when that use case is needed. (IMO the CLI should be making an API call to a running Admin UI server, that would send the email. Then the only config that would be needed would be the admin UI URL and an authentication token)

nsklikas (dismissed):
LGTM

cmd/createEmail.go
BarcoMasile: Just a small note, for the spec thing, we have this https://github.com/canonical/identity-platform-admin-ui/issues/416
in out backlog

cmd/createEmail.go
BarcoMasile: I agree the name is odd.
This email is part of the "user invite" flow, so maybe we can change this command name to reflect that? Would that help?
I was thinking some command like:
- `send-invite` or
- `user-invite`

cmd/createEmail.go
BarcoMasile: Great observations.
It would be best to invalidate that code but Kratos API don't allow for that. 
Right now we rely on the default expiration of code which is 1 hour. Which is decent, usually code like this last even longer. Also, we can improve on the security posture by shortening this time editing the `selfservice.methods.code.config.lifespan` in the kratos.yml file, so we're good!

cmd/createEmail.go
BarcoMasile: We could probably put this in a shared location under `cmd` folder. Could probably be fit into the refactoring to follow the cobra guidelines on command syntax?

BarcoMasile (approved):
LGTM, left a small suggestion for the command name change

cmd/createEmail.go
wood-push-melon: gotcha. let me change the name.

cmd/createEmail.go
wood-push-melon: gotcha. thanks for the explanation.

cmd/createEmail.go
wood-push-melon: yeah, understood your point. Had the same question when writing the code. Just followed the other code to be consistent. We can change it later for tech debt.

nsklikas (approved):

pik4ez-canonical (approved):

cmd/createEmail.go
shipperizer: logger.Fatalf is probably what we want here

PR #429, opened 2024-09-27
fix: local dev env for OIDC provider discovery

BarcoMasile:
## Description
This fixes the skaffold dev environment to allow local runs of the app.

### DIFF
--- a/skaffold.yaml
+++ b/skaffold.yaml
@@ -89,7 +89,7 @@ portForward:
   resourceName: hydra-public
   namespace: default
   port: 4444
-  localPort: 14444
+  localPort: 4444
 - resourceType: service
   resourceName: oathkeeper-api
   namespace: default


pik4ez-canonical (approved):

skaffold.yaml
```diff
@@ -89,7 +89,7 @@ portForward:
   resourceName: hydra-public
   namespace: default
   port: 4444
-  localPort: 14444
+  localPort: 4444
```
nsklikas, line 92: question: what is the problem with the current configuration? I thought that we weren't going to use hydra as a provider. 

If we want to enable hydra as a provider, why don't we change the issuer_url to point to port `14444` instead to keep consistency with how we expose apps via skaffold?

BarcoMasile: The problem is that the local setup doesn't allow the app to work, and while developing I need to have a quick way to try things without relying on Canonical infra.

The specific problem is the issuer validation, which doesn't match the one queried when the response is returned from the oidc discovery, causing the startup to fail (it would still fail validation on tokens "later" in the app life).

I changed it here because the skaffold deployment was the "broken" one, meanwhile the docker compose one was not.
If we keep the 14444 port here, we would have to change it in the docker compose (breaking consistency over there as well).
We could just adopt the same approach for all ports in both of our local deployment setups. My main focus here was fixing things so it works.
I'm open to whatever in this case

nsklikas (approved):

Issue #431, opened 2024-10-09
Investigate if basePath workaround is still needed


nsklikas:
After adding the base html tag to our index.html, we need to investigate whether the basePath workaround (https://github.com/canonical/identity-platform-admin-ui/blob/main/ui/src/util/basePaths.ts) is still needed. If not, we should remove it

PR #432, opened 2024-10-10
IAM-1137 - refactor: fetch basename from base tag

huwshimi:
Fetch the basename for the router from the `<base>` tag. React Router doesn't support the base tag by default so we have to manually get the path.

https://warthogs.atlassian.net/browse/IAM-1137
Fixes: #431.

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -10,7 +10,6 @@
 
     <script>
         const global = globalThis;
-        const base = "{{ . }}";
     </script>
     <base href="{{ . }}ui/" />
   </head>
--- a/ui/src/util/basePaths.spec.ts
+++ b/ui/src/util/basePaths.spec.ts
@@ -5,34 +5,59 @@ import {
 } from "./basePaths";
 
 vi.mock("./basePaths", async () => {
-  window.base = "/example";
+  const basePath = "/example/ui";
+  // Can't use the setBase function here as this mock gets hoisted to the top of
+  // the file so gets executed before setBase exists.
+  const base = document.createElement("base");
+  base.setAttribute("href", basePath);
+  document.body.appendChild(base);
   const actual = await vi.importActual("./basePaths");
   return {
     ...actual,
-    basePath: "/example/ui/",
-    apiBasePath: "/example/api/v0/",
+    basePath,
   };
 });
 
+const cleanBase = () =>
+  document.querySelectorAll("base").forEach((base) => {
+    document.body.removeChild(base);
+  });
+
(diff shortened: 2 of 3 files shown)


ui/index.html
```diff
@@ -10,7 +10,6 @@
 
     <script>
         const global = globalThis;
-        const base = "{{ . }}";
```
huwshimi, line 13: This is no longer required, so we have one less way to set the base.

ui/src/util/basePaths.ts
```diff
@@ -1,25 +1,18 @@
...
+    return `${removeTrailingSlash(path)}/` as BasePath;
   }
-  return "/";
+  return "/ui/";
 };
 
-export const basePath: BasePath = `${calculateBasePath()}ui`;
-export const apiBasePath: BasePath = `${calculateBasePath()}api/v0/`;
+export const basePath: BasePath = calculateBasePath();
+export const apiBasePath: BasePath = "/api/v0/";
```
huwshimi, line 15: Looking at the backed the API is always at the root and isn't prefixed by the context path like the UI is. Is that correct?

ui/src/util/basePaths.spec.ts
```diff
@@ -5,34 +5,59 @@ import {
...
 
-  it("resolves with root path if the base is not provided", () => {
-    if (window.base) {
-      delete window.base;
-    }
+  //   it("handles full URL", () => {
+  //     setBase("http://example.com/test/");
+  //     const result = calculateBasePath();
+  //     expect(result).toBe("/test/");
+  //   });
```
vladimir-cucu, line 56: This might need to be removed or uncommented.

huwshimi: Oops! Thanks.

ui/src/util/basePaths.ts
nsklikas: No, when running behind a proxy the backend may be served under a path. This is how our charmed traefik works.

The only thing you know is that the backend is served on `{basePath}/..` (because `basepath == "https://path/to/backend/ui`). On top of that there is the convention that all our handlers are served under `api`

BarcoMasile: Exactly, it is always been like that.
To sum up
You can't run the admin-ui executable APIs under a context path (on local machine for example).
But when deploying, traefik proxy will handle the request under a context path, letting admin ui "believe" it's under no path at all.

shipperizer (approved):

PR #443, opened 2024-10-22
refactor: use rebac admin package for admin-ui

huwshimi:
## Changes

- Replace rebac admin package with admin-ui specific package (the main package will contain the v1 API ReBAC Admin client).
- Fix template for local development.

https://warthogs.atlassian.net/browse/WD-16023

### DIFF
--- a/ui/index.html
+++ b/ui/index.html
@@ -6,17 +6,23 @@
     <meta name="viewport" content="width=device-width, initial-scale=1" />
 
     <title>Identity platform</title>
-    <link rel="shortcut icon" href="https://assets.ubuntu.com/v1/49a1a858-favicon-32x32.png" type="image/x-icon" />
+    <link
+      rel="shortcut icon"
+      href="https://assets.ubuntu.com/v1/49a1a858-favicon-32x32.png"
+      type="image/x-icon"
+    />
 
     <script>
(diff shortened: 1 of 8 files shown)


ui/index.html
```diff
@@ -6,17 +6,23 @@
...
 
     <script>
-        const global = globalThis;
+      const global = globalThis;
     </script>
+    <% if(process.env.NODE_ENV === "development"){ %>
+    <base href="/ui/" />
+    <% } else{ %>
     <base href="{{ . }}ui/" />
+    <% } %>
```
BarcoMasile, line 22: The way we populate this value already takes care of normalizing the path for the slashes, so this is not needed

nsklikas (approved):

ui/index.html
huwshimi: Sorry, which part do you mean isn't needed? This change is for development where the UI is run with Vite and so the `{{ . }}` is not transformed to a value (so the base literally becomes `{{ . }}ui/`).

vladimir-cucu (approved):

ui/index.html
BarcoMasile: That's ok, then I'll resolve this. Thank you!

Issue #455, opened 2024-11-12
Improve both communication and implementation of the Mapper choice for Provider creation


BarcoMasile:
## Description
When trying to create a Provider in the Admin UI frontend, you have an option to specify a Mapper.
This could be either a:
- URL (`https://`, `file://` or `base64://`)
- inline JSON Snippet

We need to perform some changes.

### First step
The description for the field needs to reflect this on the frontend.
The backend needs to make sure it is able to handle all those value types.

### Second step
We need to expose an endpoint to allow the frontend to populate a simple dropdown with available mappers.
The frontend needs to allow the user to perform a choice between one of the available providers or to specify a custom info like it's supposed to do now.

![image](https://github.com/user-attachments/assets/5f004924-3592-4a3c-86eb-8773961a920c)

## References
Related to https://github.com/canonical/identity-platform-admin-ui/issues/405

Issue #405, opened 2024-09-10
Provider creation defaults to non-existant schema


nsklikas:
When trying to create a provider, I am asked to input the `mapper`:
![image](https://github.com/user-attachments/assets/9b74e8b3-e36f-4bd7-8384-f53c75dda460)

`file:///etc/config/kratos/okta_schema.jsonnet` is hardcoded into the code, but this file does not exist in the charm. Should the backend implement an API to expose the available schemas?

PR #456, opened 2024-11-12
IAM-1047 provider defaults to non existant schema

BarcoMasile:
## Description
The only way to address https://github.com/canonical/identity-platform-admin-ui/issues/405 right now is to remove the default value.

I opened issue https://github.com/canonical/identity-platform-admin-ui/issues/455 to address the lack of a dropdown and more info about how to populate the Mapper field in the Create Provider form.

## Changes
- Since Kratos supports OIDC discovery, for the `generic` provider type we are allowing AuthURL and TokenURL to be empty in case IssuerURL is non empty
- Also, now IssuerURL can be empty if AuthURL and TokenURL are both provided
- Mapper is also allowed to be empty since Kratos will use the default schema in case that field is empty
- added a custom validation function to validate AuthURL and TokenURL presence in case IssuerURL is not specified
- added test entry to test the changes

### DIFF
--- a/pkg/idp/third_party.go
+++ b/pkg/idp/third_party.go
@@ -50,19 +50,19 @@ type Configuration struct {
 	// IssuerURL is the OpenID Connect Server URL. You can leave this empty if `provider` is not set to `generic`.
 	// If set, neither `auth_url` nor `token_url` are required.
 	// validate that this field is required only when Provider field == "generic"
-	IssuerURL string `json:"issuer_url" yaml:"issuer_url" validate:"required_if=Provider generic"`
+	IssuerURL string `json:"issuer_url" yaml:"issuer_url"`
 
 	// AuthURL is the authorize url, typically something like: https://example.org/oauth2/auth
 	// Should only be used when the OAuth2 / OpenID Connect server is not supporting OpenID Connect Discovery and when
 	// `provider` is set to `generic`.
-	// validate that this field is required only when Provider field == "generic"
-	AuthURL string `json:"auth_url" yaml:"auth_url" validate:"required_if=Provider generic"`
+	// validate that this field is required only when Provider field == "generic" and IssuerURL is empty
+	AuthURL string `json:"auth_url" yaml:"auth_url"`
 
 	// TokenURL is the token url, typically something like: https://example.org/oauth2/token
 	// Should only be used when the OAuth2 / OpenID Connect server is not supporting OpenID Connect Discovery and when
 	// `provider` is set to `generic`.
-	// validate that this field is required only when Provider field == "generic"
-	TokenURL string `json:"token_url" yaml:"token_url" validate:"required_if=Provider generic"`
+	// validate that this field is required only when Provider field == "generic" and IssuerURL is empty
+	TokenURL string `json:"token_url" yaml:"token_url"`
 
 	// Tenant is the Azure AD Tenant to use for authentication, and must be set when `provider` is set to `microsoft`.
 	// Can be either `common`, `organizations`, `consumers` for a multitenant application or a specific tenant like
@@ -103,7 +103,7 @@ type Configuration struct {
 	// profile information) to hydrate the identity's data.
 	//
 	// It can be either a URL (file://, http(s)://, base64://) or an inline JSONNet code snippet.
-	Mapper string `json:"mapper_url" yaml:"mapper_url" validate:"required"`
+	Mapper string `json:"mapper_url" yaml:"mapper_url"`
 
 	// RequestedClaims string encoded json object that specifies claims and optionally their properties which should be
 	// included in the id_token or returned from the UserInfo Endpoint.
--- a/pkg/idp/validation.go
+++ b/pkg/idp/validation.go
@@ -23,9 +23,24 @@ type PayloadValidator struct {
 	logger logging.LoggerInterface
 }
 
+func genericIssuerOAuth2URLsValidation(sl validator.StructLevel) {
+	configuration := sl.Current().Interface().(Configuration)
+
+	if configuration.Provider != "generic" {
+		return
+	}
+
+	// Kratos will try OIDC discovery, so if IssuerURL is not empty, AuthURL and TokenURL could be empty
+	// if IssuerURL is empty, then we need both AuthURL and TokenURL
+	if configuration.IssuerURL == "" && (configuration.AuthURL == "" || configuration.TokenURL == "") {
+		sl.ReportError(configuration.IssuerURL, "issuer_url", "IssuerURL", "issuer_urls", "")
+	}
+}
+
 func (p *PayloadValidator) setupValidator() {
 	// validate Provider to be one of the supported ones
 	p.validator.RegisterAlias("supported_provider", fmt.Sprintf("oneof=%s", SUPPORTED_PROVIDERS))
+	p.validator.RegisterStructValidation(genericIssuerOAuth2URLsValidation, Configuration{})
 }
 
 func (p *PayloadValidator) NeedsValidation(r *http.Request) bool {
--- a/pkg/idp/validation_test.go
+++ b/pkg/idp/validation_test.go
@@ -120,6 +120,45 @@ func TestValidate(t *testing.T) {
 			expectedError:  nil,
 		},
 		{
+			name:     "CreateIdPSuccessWithOIDCDiscovery",
+			method:   http.MethodPost,
+			endpoint: "",
+			body: func() []byte {
+				conf := new(Configuration)
+				conf.ID = "google_generic"
+				conf.Provider = "generic"
+				conf.IssuerURL = "mock-url"
+				conf.SubjectSource = "me"
+				conf.Scope = []string{}
+				conf.Mapper = "mock-url"
+
+				marshal, _ := json.Marshal(conf)
+				return marshal
+			},
+			expectedResult: nil,
+			expectedError:  nil,
+		},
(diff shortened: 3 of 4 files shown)


pkg/idp/third_party.go
```diff
@@ -55,14 +55,14 @@ type Configuration struct {
 	// AuthURL is the authorize url, typically something like: https://example.org/oauth2/auth
 	// Should only be used when the OAuth2 / OpenID Connect server is not supporting OpenID Connect Discovery and when
 	// `provider` is set to `generic`.
-	// validate that this field is required only when Provider field == "generic"
-	AuthURL string `json:"auth_url" yaml:"auth_url" validate:"required_if=Provider generic"`
+	// validate that this field is required only when Provider field == "generic" and IssuerURL is not empty
```
nsklikas, line 58: Shouldn't this be `when Provider field == "generic" and IssuerURL is empty` (if issuerURL is not empty then we can (?) assume that discovery is supported

pkg/idp/third_party.go
```diff
@@ -55,14 +55,14 @@ type Configuration struct {
...
+	AuthURL string `json:"auth_url" yaml:"auth_url" validate:"required_if=IssuerURL;required_if=Provider generic"`
```
nsklikas, line 59: Shouldn't this be:
```suggestion
	AuthURL string `json:"auth_url" yaml:"auth_url" validate:"required_unless=IssuerURL,required_if=Provider generic"`
```

AFAICT to chain validators you need `,` (see [here](https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Multiple_Validators)). Also we should require the AuthURL is `IssuerURL` is empty (not the other way around).

I haven't tested this so please correct me.

pkg/idp/validation_test.go
nsklikas, line 1: Could we add tests for:
- empty issuerURL and empty `AuthURL`/`TokenURL` (should error)
- empty issuerURL and not empty `AuthURL`/`TokenURL` (should succeed)

nsklikas (commented):
Haven't tried it, so I am probably wrong on some things

pkg/idp/third_party.go
BarcoMasile: I got confused, you're 100% right. Fixing it

pkg/idp/third_party.go
BarcoMasile: Got confused here also, you're right

pkg/idp/third_party.go
```diff
@@ -55,14 +55,14 @@ type Configuration struct {
...
+	// validate that this field is required only when Provider field == "generic" and IssuerURL is empty
+	AuthURL string `json:"auth_url" yaml:"auth_url" validate:"required_unless=IssuerURL,required_if=Provider generic"`
 
 	// TokenURL is the token url, typically something like: https://example.org/oauth2/token
 	// Should only be used when the OAuth2 / OpenID Connect server is not supporting OpenID Connect Discovery and when
 	// `provider` is set to `generic`.
-	// validate that this field is required only when Provider field == "generic"
-	TokenURL string `json:"token_url" yaml:"token_url" validate:"required_if=Provider generic"`
+	// validate that this field is required only when Provider field == "generic" and IssuerURL is empty
+	TokenURL string `json:"token_url" yaml:"token_url" validate:"required_unless=IssuerURL,required_if=Provider generic"`
```
pik4ez-canonical, line 65: I believe [`required_without`](https://pkg.go.dev/github.com/go-playground/validator/v10#hdr-Required_Without) is the validation rule that should be used instead of `required_unless`. The `required_unless` rule [needs](https://github.com/canonical/identity-platform-admin-ui/actions/runs/11801933476/job/32876428929?pr=456#step:6:40) an exact value of the other field.

Unfortunately, there might still be a logical issue here. Let's take a look at the following example:
```
IssuerURL: some.issuer/url
AuthURL: nil
TokenURL: nil
Provider: generic
```

Having the TokenURL validation rule `required_without=IssuerURL,required_if=Provider generic`:
* `required_without=IssuerURL` check will pass. TokenURL isn't required, as IssuerURL is set.
* `required_if=Provider generic` check will fail. TokenURL is always required when Provider set to `generic`.

My domain knowledge is limited. But from the first glance, the flexibility of validation rules might not be enough for our case. Custom validation implementation might be an answer.

pik4ez-canonical (changes requested):

pkg/idp/third_party.go
BarcoMasile: FYI @nsklikas

nsklikas: I have not tried it, but I don't understand it. In this case:

 ```
    IssuerURL: some.issuer/url 
    AuthURL: nil
    TokenURL: nil
    Provider: generic
```
> Having the TokenURL validation rule required_without=IssuerURL,required_if=Provider generic:
    - required_without=IssuerURL check will pass. TokenURL isn't required, as IssuerURL is set.
    - required_if=Provider generic check will fail. TokenURL is always required when Provider set to generic.

`required_without=IssuerURL` should fail, so the second rule shouldn't be considered. This means that it should be evaluated as required. From reading the docs I would expect that chaining conditions would happen as an `AND` not an `OR`. Am I missing something?

Comment (nsklikas) [reply pkg/idp/third_party.go:65]: If it turns out too complicated, let's park it and move it to a different PR.

Comment (BarcoMasile) [reply pkg/idp/third_party.go:65]: It works as an AND. I'm experimenting with a or also (`|`) but still something is off.
I'll check the custom validation func way

File: pkg/idp/validation.go [hunk 4]
```diff
@@ -23,9 +23,22 @@ type PayloadValidator struct {
 	logger logging.LoggerInterface
 }
 
+func genericIssuerOAuth2URLsValidation(sl validator.StructLevel) {
```
BarcoMasile, line 26: Note: long name, but I couldn't find a better one at explaining what this does.

pkg/idp/validation.go
```diff
@@ -23,9 +23,22 @@ type PayloadValidator struct {
...
+	configuration := sl.Current().Interface().(Configuration)
+
+	if configuration.Provider == "generic" {
```
wood-push-melon, line 29: Just a minor nitpick (feel free to ignore):

We could return early to do one less `if` block:

```go
if configuration.Provider != "generic" {
    return
}

if configuration.IssuerURL == "" && (configuration.AuthURL == "" || configuration.TokenURL == "") {
		sl.ReportError(configuration.IssuerURL, "issuer_url", "IssuerURL", "issuer_urls", "")
}
```

wood-push-melon (dismissed):

pkg/idp/validation.go
BarcoMasile: Done :) Thanks for the suggestion! I pushed again with the change

pik4ez-canonical (approved):
Looks clean and readable thanks to the comments.

Thank you Marco, good job!

pkg/idp/validation_test.go
```diff
@@ -119,6 +119,46 @@ func TestValidate(t *testing.T) {
...
+		},
+		{
+			name:     "CreateIdPSuccessWithoutOIDCDiscovery",
+			method:   http.MethodPost,
+			endpoint: "",
+			body: func() []byte {
+				conf := new(Configuration)
+				conf.ID = "google_generic"
+				conf.Provider = "generic"
+				conf.IssuerURL = "mock-url"
```
nsklikas, line 149: `IssuerURL` should be empty

pkg/idp/third_party.go
```diff
@@ -55,14 +55,14 @@ type Configuration struct {
 	// AuthURL is the authorize url, typically something like: https://example.org/oauth2/auth
 	// Should only be used when the OAuth2 / OpenID Connect server is not supporting OpenID Connect Discovery and when
 	// `provider` is set to `generic`.
-	// validate that this field is required only when Provider field == "generic"
-	AuthURL string `json:"auth_url" yaml:"auth_url" validate:"required_if=Provider generic"`
+	// validate that this field is required only when Provider field == "generic" and IssuerURL is empty
+	AuthURL string `json:"auth_url" yaml:"auth_url"`
```
nsklikas, line 59: IMHO we don't need the custom validator. This seems to work:
```
AuthURL string `json:"auth_url" yaml:"auth_url" validate:"excluded_with=IssuerURL&&required_if=Provider generic"`
```

BarcoMasile: The `excluded_with` doesn't appear in the linked sections of the doc like the other tags, so I did not see that.
I don't see documentation for it. Also, the && is a comma or I should test the actual ampersand?

BarcoMasile: I tested your solution, I don't like the use of the ampersand because it is not supported by the library.
The `and` is expressed through the comma `,`, and with that it doesn't work.
Based on these findings, I have to stick with the custom function (as much as I don't fancy it).

pkg/idp/validation_test.go
BarcoMasile: I removed the requirement on the IssuerURL so it can actually be empty in case authurl and tokenurl are present, for the generic provider. I'm adjusting the test right now.

BarcoMasile:
I consider this PR final, all comments are addressed and already rebased on main, please review again @nsklikas @pik4ez-canonical

pik4ez-canonical (approved):

nsklikas (approved):

PR #473, opened 2024-11-21
Fix skaffold setup

nsklikas:
Fix based on https://github.com/openfga/helm-charts/issues/161


shipperizer (approved):
we should also pin down the chart version eventually

natalian98 (approved):

PR #485, opened 2024-11-27
WD-17050 - chore(tests): add Navigation component tests

huwshimi:
## Changes

Add tests for the Navigation component.

https://warthogs.atlassian.net/browse/WD-17050

### DIFF
--- a/ui/src/components/Navigation/Navigation.test.tsx
+++ b/ui/src/components/Navigation/Navigation.test.tsx
@@ -1,0 +1,58 @@
+import { renderComponent } from "test/utils";
+import { screen, within } from "@testing-library/dom";
+
+import { urls } from "urls";
+
+import Navigation from "./Navigation";
+import { Label } from "./types";
+import { authURLs } from "api/auth";
+import { appendAPIBasePath } from "util/basePaths";
+
+test("displays page links", () => {
+  renderComponent(<Navigation />);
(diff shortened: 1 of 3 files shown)


ui/src/components/Navigation/Navigation.test.tsx
```diff
@@ -0,0 +1,58 @@
...
+  );
+  expect(screen.getByRole("link", { name: Label.IDENTITIES })).toHaveAttribute(
+    "href",
+    urls.identities.index,
+  );
+  expect(screen.getByRole("link", { name: Label.SCHEMAS })).toHaveAttribute(
+    "href",
+    urls.schemas.index,
+  );
+  expect(screen.getByRole("link", { name: "Groups" })).toHaveAttribute(
```
BarcoMasile, line 28: question: why not adding `"Groups"` and `"Roles"` to the labels as well?

nsklikas (approved):

ui/src/components/Navigation/Navigation.test.tsx
huwshimi: The link labels are defined inside the external ReBAC Admin package so these strings are only used in a single place in this codebase (these tests), but I can add them to the enum as well if you think that would be better?

BarcoMasile (approved):

PR #487, opened 2024-11-28
WD-17053 - chore(test): add Panel tests

huwshimi:
## Changes

Add tests for the Panel component.

https://warthogs.atlassian.net/browse/WD-17053


nsklikas (dismissed):

ui/src/pages/clients/ClientCreate/types.ts
BarcoMasile, line 1: suggestion: how about moving test related entities to an explicitly test-bound file?

BarcoMasile: I see this is useful for the `data-testid` attribute. 
I'm not a huge a fan of mixing non-test code with test-related utilities,
but I understand this is useful. I would just have the enum in a strictly test-related file

huwshimi: I've moved these types to `test-types.ts` files. I've also added a util that only includes the test ids when it's not a production build.

BarcoMasile: @huwshimi That's more that I could ask for, thank you very much!
I see that the util `testId` you provided is only used in `SidePanel.tsx`
Would you adopt it in the other spots in this PR also or do you plan to do it opportunistically in subsequent PRs ?

huwshimi: Oops thanks, I've added it for all of them.

BarcoMasile (approved):

nsklikas (approved):

PR #495, opened 2024-12-03
WD-17060 - chore(tests): add delete button tests

huwshimi:
## Changes

Create shared delete button component.
Add tests to delete buttons.

https://warthogs.atlassian.net/browse/WD-17060
https://warthogs.atlassian.net/browse/WD-17062
https://warthogs.atlassian.net/browse/WD-17066

### DIFF
--- a/ui/src/components/DeletePanelButton/DeletePanelButton.tsx
+++ b/ui/src/components/DeletePanelButton/DeletePanelButton.tsx
@@ -1,0 +1,62 @@
+import { FC, useState } from "react";
+import { useNavigate } from "react-router-dom";
+import { useQueryClient } from "@tanstack/react-query";
+import { ConfirmationButton, useNotify } from "@canonical/react-components";
+import { Label, Props } from "./types";
+
+const DeletePanelButton: FC<Props> = ({
+  confirmButtonDisabled,
+  confirmButtonLabel,
+  confirmContent,
+  confirmTitle = "Confirm delete",
+  invalidateQuery,
+  entityName,
+  onDelete,
+  successMessage,
+  successPath,
+}) => {
+  const notify = useNotify();
+  const queryClient = useQueryClient();
+  const [isLoading, setLoading] = useState(false);
+  const navigate = useNavigate();
+
+  const handleDelete = () => {
+    setLoading(true);
+    onDelete()
+      .then(() => {
+        navigate(successPath, notify.queue(notify.success(successMessage)));
+      })
+      .catch((error: unknown) => {
+        notify.failure(
+          `${entityName} deletion failed`,
+          error instanceof Error ? error : null,
+          typeof error === "string" ? error : null,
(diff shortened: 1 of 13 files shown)


ui/src/components/DeletePanelButton/DeletePanelButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const navigate = useNavigate();
+
+  const handleDelete = () => {
+    setLoading(true);
+    onDelete()
+      .then(() => {
+        navigate(successPath, notify.queue(notify.success(successMessage)));
+      })
+      .catch((e) => {
+        notify.failure(`${entityName} deletion failed`, e);
```
BarcoMasile, line 30: question: is the `e` error type human readable enough for it to be displayed in the notification?

BarcoMasile (dismissed):

ui/src/components/DeletePanelButton/DeletePanelButton.tsx
huwshimi: That's a good question. This will display whatever is returned in the error `Message` param from the API, which at a quick look at the backend appears to be whatever internal error might occur, is that correct? If so I'll update this to not display those messages.

BarcoMasile: In that case I would expect the template to be something like this
```
notify.failure(`${entityName} deletion failed`, e.Message ?? e);
```

nsklikas (dismissed):

ui/src/components/DeletePanelButton/DeletePanelButton.tsx
huwshimi: The message gets extracted by the API handler (https://github.com/canonical/identity-platform-admin-ui/blob/bffb816465023abfe249b21615b8bfdea5509848/ui/src/util/api.ts#L14 - lower case m from "Message string `json:"message"`" in the backend).

And the second param of `notify.failure()` takes an `Error` object which it then uses to extract the message: https://github.com/canonical/react-components/blob/83acef1f633fcf1a73641dca4287b8d89d944b96/src/components/NotificationProvider/messageBuilder.tsx#L41-L44.

However it looks like there was an issue with our code as depending on where the error comes from it might be an `Error` object or string so I've done some additional handling here to pass the message to the third param of `notify.failure()` if it's a string and added an extra test.

BarcoMasile: Thank you @huwshimi , nice work!

BarcoMasile:
@huwshimi if you can, try squashing those merge commits, we try to avoid that whenever possible. Thanks

huwshimi:
> @huwshimi if you can, try squashing those merge commits, we try to avoid that whenever possible. Thanks

Done!

PR #513, opened 2025-01-16
Upgrade openfga dependency to support ListUsers API

BarcoMasile:
## Description
Upgraded github.com/openfga/go-sdk from v0.3.4 to v0.3.7 (which is the last release that does not introduce breaking changes while at same time supporting the new ListUsers API we need).


BarcoMasile:
> > we might need to signal that to renovate or have a specific test that fails if the lib is updated beyond 0.3.7
> 
> @shipperizer I can look into it

Here's the PR https://github.com/canonical/identity-platform-admin-ui/pull/514

natalian98 (approved):

nsklikas (approved):

PR #535, opened 2025-02-05
IAM-1285 Adopt ListUsers API

BarcoMasile:
## Description
This PR build on the preparatory work of the previous OpenFGA upgrades from the PRs mentioned below.
Adoption of the ListUsers API in both the `roles` package and the `groups` package to improve the implementation of the 
- `roles.ListRoleGroups`
- `groups.ListGroupIdentities`

The `OpenFGACoreClientInterface` for the core OpenFGA client has been updated with the additional methods for the ListUsers API.

The `ListUsers` replaces the `ReadTuples`. One difference to note is that the ListUsers doesn't provide paginated results so the continuation token has been removed from the signature and the return type.

Tests have been adjusted accordingly.

## References
- https://github.com/canonical/identity-platform-admin-ui/pull/513
- https://github.com/canonical/identity-platform-login-ui/pull/427
- https://github.com/canonical/openfga-operator/pull/146
- https://github.com/canonical/openfga-rock/pull/15
- https://github.com/canonical/identity-platform-admin-ui/pull/514

### DIFF
--- a/internal/openfga/client.go
+++ b/internal/openfga/client.go
@@ -1,5 +1,5 @@
-// Copyright 2024 Canonical Ltd
-// SPDX-License-Identifier: AGPL
+// Copyright 2025 Canonical Ltd.
+// SPDX-License-Identifier: AGPL-3.0
 
 package openfga
 
@@ -354,6 +354,58 @@ func (c *Client) ListObjects(ctx context.Context, user, relation, objectType str
 	for i, p := range objectsResponse.GetObjects() {
 		// remove the "{objectType}:" prefix from the response
 		allowedObjs[i] = p[len(fmt.Sprintf("%s:", objectType)):]
+	}
+
+	return allowedObjs, nil
+}
+
+func (c *Client) ListUsers(ctx context.Context, userFilter, relation, object string) ([]string, error) {
+	ctx, span := c.tracer.Start(ctx, "openfga.Client.ListUsers")
+	defer span.End()
+
+	var filter openfga.UserTypeFilter
+
+	userType, userRelation, _ := strings.Cut(userFilter, "#")
+
+	filter = openfga.UserTypeFilter{
+		Type:     userType,
+		Relation: &userRelation,
+	}
+
+	objectType, objectID, _ := strings.Cut(object, ":")
+
+	listUsersReq := c.c.ListUsers(ctx)
+
+	listUsersReq = listUsersReq.Body(client.ClientListUsersRequest{
+		Object:      openfga.FgaObject{Type: objectType, Id: objectID},
+		Relation:    relation,
+		UserFilters: []openfga.UserTypeFilter{filter},
+	})
+
+	usersResponse, err := c.c.ListUsersExecute(listUsersReq)
+	if err != nil {
+		c.logger.Errorf("issues performing list users operation: %s", err)
+		return nil, err
+	}
+
+	users := usersResponse.GetUsers()
+
+	allowedObjs := make([]string, len(users))
+	for i, user := range users {
+		if o, ok := user.GetObjectOk(); ok {
+			allowedObjs[i] = fmt.Sprintf("%s:%s", o.GetType(), o.GetId())
+			continue
+		}
+
+		if us, ok := user.GetUsersetOk(); ok {
+			allowedObjs[i] = fmt.Sprintf("%s:%s", us.GetType(), us.GetId())
+			continue
+		}
+
+		if u, ok := user.GetWildcardOk(); ok {
+			allowedObjs[i] = fmt.Sprintf("%s:*", u.GetType())
+			continue
+		}
 	}
 
 	return allowedObjs, nil
--- a/pkg/groups/service.go
+++ b/pkg/groups/service.go
@@ -1,4 +1,4 @@
-// Copyright 2024 Canonical Ltd.
+// Copyright 2025 Canonical Ltd.
 // SPDX-License-Identifier: AGPL-3.0
 
 package groups
@@ -357,30 +357,24 @@ func (s *Service) DeleteGroup(ctx context.Context, ID string) error {
 	return nil
 }
 
-// ListIdentities returns all the identities (users for now) assigned to a group
-func (s *Service) ListIdentities(ctx context.Context, ID, continuationToken string) ([]string, string, error) {
+// ListIdentities returns all the identities (users and group#member associated) assigned to a group
+func (s *Service) ListIdentities(ctx context.Context, ID string) ([]string, error) {
 	ctx, span := s.tracer.Start(ctx, "groups.Service.ListIdentities")
 	defer span.End()
 
-	r, err := s.ofga.ReadTuples(ctx, "", authz.MEMBER_RELATION, authz.GroupForTuple(ID), continuationToken)
-
+	users, err := s.ofga.ListUsers(ctx, "user", authz.MEMBER_RELATION, authz.GroupForTuple(ID))
 	if err != nil {
 		s.logger.Error(err.Error())
-		return nil, "", err
+		return nil, err
 	}
 
-	identities := make([]string, 0)
-
-	for _, t := range r.GetTuples() {
-		// TODO @shipperizer the user: bit will have to change when or if we use the identity type, this will be tricky
-		// best way right now might be to verify if a user is also an identity (no idea how though)
-		// at the moment an identity cannot be a member of a group, only a user
-		if strings.HasPrefix(t.Key.User, "user:") {
-			identities = append(identities, t.Key.User)
-		}
+	groups, err := s.ofga.ListUsers(ctx, "group#member", authz.MEMBER_RELATION, authz.GroupForTuple(ID))
+	if err != nil {
+		s.logger.Error(err.Error())
+		return nil, err
 	}
 
-	return identities, r.GetContinuationToken(), nil
+	return append(users, groups...), nil
 }
 
 // AssignIdentities assigns identities to a group, right now using the type user which is disconnected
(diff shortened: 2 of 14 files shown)


pkg/roles/service.go
```diff
@@ -63,26 +63,18 @@ func (s *Service) ListRoles(ctx context.Context, userID string) ([]string, error
 // TODO @shipperizer a more complex pagination system can be implemented by keeping track of the
 // latest index in the current "page" and encode it in the pagination token header returned to
 // the UI
-func (s *Service) ListRoleGroups(ctx context.Context, ID, continuationToken string) ([]string, string, error) {
+func (s *Service) ListRoleGroups(ctx context.Context, ID string) ([]string, error) {
```
wood-push-melon, line 66: just a question:

do we also need to update the comment above?

pkg/roles/handlers.go
```diff
@@ -317,17 +317,7 @@ func (a *API) handleListRoleGroup(w http.ResponseWriter, r *http.Request) {
...
-	if err := paginator.LoadFromRequest(r.Context(), r); err != nil {
-		a.logger.Error(err)
-	}
-
-	roles, pageToken, err := a.service.ListRoleGroups(
-		r.Context(),
-		ID,
-		paginator.GetToken(r.Context(), ROLE_TOKEN_KEY),
-	)
+	roles, err := a.service.ListRoleGroups(r.Context(), ID)
```
wood-push-melon, line 320: `roles` -> `groups`?

pkg/groups/service.go
```diff
@@ -358,29 +363,24 @@ func (s *Service) DeleteGroup(ctx context.Context, ID string) error {
...
 
 // ListIdentities returns all the identities (users for now) assigned to a group
-func (s *Service) ListIdentities(ctx context.Context, ID, continuationToken string) ([]string, string, error) {
+func (s *Service) ListIdentities(ctx context.Context, ID string) (*ListIdentitiesResult, error) {
 	ctx, span := s.tracer.Start(ctx, "groups.Service.ListIdentities")
 	defer span.End()
 
-	r, err := s.ofga.ReadTuples(ctx, "", authz.MEMBER_RELATION, authz.GroupForTuple(ID), continuationToken)
+	users, errUsers := s.ofga.ListUsers(ctx, "user", authz.MEMBER_RELATION, authz.GroupForTuple(ID))
+	groups, errGroups := s.ofga.ListUsers(ctx, "group", authz.MEMBER_RELATION, authz.GroupForTuple(ID))
```
wood-push-melon, line 371: just a random dummy question: is it possible to merge these two calls into one?

Another thing is that:
Is it better if we do an early return if `errUsers` is not `nil` so that we can avoid the second call?

internal/openfga/client.go
```diff
@@ -359,6 +359,67 @@ func (c *Client) ListObjects(ctx context.Context, user, relation, objectType str
...
+	ctx, span := c.tracer.Start(ctx, "openfga.Client.ListUsers")
+	defer span.End()
+
+	var filter openfga.UserTypeFilter
+	var userRelation *string = nil
+
+	parts := strings.SplitN(userFilter, "#", 2)
+	userType := parts[0]
+
+	if len(parts) == 2 {
```
wood-push-melon, line 372: non-blocking: `strings.Cut` could be better?

pkg/groups/service.go
BarcoMasile: Unfortunately it's not possible to merge the calls because the ListUsers API requires the first argument, and we can only specify one at a time.
But yes, I will change the code and do the early return, thanks :)

BarcoMasile: Done in latest push!

internal/openfga/client.go
BarcoMasile: You know what, I thought it didn't make much of a difference,  but I tried it out and man it makes the code super clean! I love it, thanks for the suggestion, I didn't know about this method
It's in the latest push, thank you!

pkg/roles/service.go
BarcoMasile: Yes, thanks for spotting it

pkg/roles/handlers.go
BarcoMasile: It's about listing the multiple groups of a single role, that's why the name is like that (if that's what you meant)

pkg/roles/service.go
BarcoMasile: Done :) I updated it also in the other service

pkg/groups/service.go
```diff
@@ -357,30 +362,24 @@ func (s *Service) DeleteGroup(ctx context.Context, ID string) error {
...
-	identities := make([]string, 0)
-
-	for _, t := range r.GetTuples() {
-		// TODO @shipperizer the user: bit will have to change when or if we use the identity type, this will be tricky
-		// best way right now might be to verify if a user is also an identity (no idea how though)
-		// at the moment an identity cannot be a member of a group, only a user
-		if strings.HasPrefix(t.Key.User, "user:") {
-			identities = append(identities, t.Key.User)
-		}
+	groups, err := s.ofga.ListUsers(ctx, "group#member", authz.MEMBER_RELATION, authz.GroupForTuple(ID))
```
shipperizer, line 376: ok i mulled it over a bit, i d be more inclined to stick with one `[]string` made of identities and groups as before for the time being

doesn't trigger a ui change and from a method signature us more inline with the others

BarcoMasile: Even though I do not agree with this, after an offline sync we will address this and postpone changes to the API

nsklikas (commented):
LGTM overall. I have a couple of more generic questions:
- I have little context here, but I am curious whether we should address some of the limitations of the ListUsers API (e.g. it works on best effort, should we expose new config in the openfga charm, etc). Also:
- I think you should add a noop function for `listUsers`
- When should ReadTuples be used vs List{Objects,Users}? Going through the documentation I think both would work for our use case (but probably I am missing something)

internal/openfga/client.go
```diff
@@ -359,6 +359,67 @@ func (c *Client) ListObjects(ctx context.Context, user, relation, objectType str
...
+			allowedObjs[i] = o.Id
+		}
+
+		if us, ok := user.GetUsersetOk(); ok {
+			allowedObjs[i] = us.Id
+		}
+
+		if _, ok := user.GetWildcardOk(); ok {
+			allowedObjs[i] = "*"
+		}
```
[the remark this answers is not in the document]
BarcoMasile: Nice one, thank you, I'll do it in next push

internal/openfga/client.go
```diff
@@ -359,6 +359,67 @@ func (c *Client) ListObjects(ctx context.Context, user, relation, objectType str
...
+
+	users := usersResponse.GetUsers()
+
+	allowedObjs := make([]string, len(users))
+	for i, user := range users {
+		if o, ok := user.GetObjectOk(); ok {
+			allowedObjs[i] = o.Id
+		}
+
+		if us, ok := user.GetUsersetOk(); ok {
```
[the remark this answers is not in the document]
BarcoMasile: The Id will be `*` so it will not have the `#member` part. 
In the Groups and Roles endpoint that rely on the ListUsers right now the current authorization model doesn't allow for that so this will serve for the case in which the ListUsers is used with other resources as well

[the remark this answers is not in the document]
BarcoMasile: I should also say that I'm working on addressing https://github.com/canonical/identity-platform-admin-ui/pull/535#discussion_r1944529444 so with next push it will actually be "object-type:object-id"

BarcoMasile:
@nsklikas 
> LGTM overall. I have a couple of more generic questions:
> 
> * I have little context here, but I am curious whether we should address some of the limitations of the ListUsers API (e.g. it works on best effort, should we expose new config in the openfga charm, etc). Also:
> * I think you should add a noop function for `listUsers`
> * When should ReadTuples be used vs List{Objects,Users}? Going through the documentation I think both would work for our use case (but probably I am missing something)

The previous task before starting these set of PRs to upgrade, was for evaluating the choice. The only downside is losing pagination but the tradeoff is worth it.
ReadTuples forces us to "manually" do filtering on the object types after they were returned by OpenFGA which is the main reason why we wanted to swap it out.

I don't think other changes are needed, since the upgrade of the OpenFGA library are already done in the previous PRs listed in the description of this PR.

I'll look into adding the noop fuction right away, thanks!

nsklikas:
I was mainly referring to this https://openfga.dev/docs/interacting/relationship-queries#caveats-1:
> ListUsers will return the results found within the time allotted (listUsersDeadline, default: 3s) up to the maximum number of results configured (listUsersMaxResults, default: 1000). See [Configuring the Server](https://openfga.dev/docs/getting-started/setup-openfga/configure-openfga)) for more on how to change the default configuration.
    - If you set listUsersDeadline to 1s, the server will spend at most 1 second finding results.
    - If you set listUsersMaxResults to 10, the server will return, at most, 10 objects.
If the number of users matching that filter is high, you should set a high value for listUsersDeadline. If the number of users matching that filter that could have that relation with the object is high, you should set a high value for listUsersMaxResults.

If I understand correctly, this means that if we don' change the default config this will return at max 1000 users. TBH this is a large enough number, but there may be cases where we have more users (?). My question is if we need to expose it in the charm to allow for such cases or if we need to warn the user somehow.

BarcoMasile:
> If I understand correctly, this means that if we don' change the default config this will return at max 1000 users. TBH this is a large enough number, but there may be cases where we have more users (?). My question is if we need to expose it in the charm to allow for such cases or if we need to warn the user somehow.

Oh yes, I think we should be good with the default values for now.
But I don't have anything against introducing additional config in the charm
cc @shipperizer

nsklikas (approved):

shipperizer:
> > If I understand correctly, this means that if we don' change the default config this will return at max 1000 users. TBH this is a large enough number, but there may be cases where we have more users (?). My question is if we need to expose it in the charm to allow for such cases or if we need to warn the user somehow.
> 
> Oh yes, I think we should be good with the default values for now.
> But I don't have anything against introducing additional config in the charm
> cc @shipperizer 


I'm afraid this is the main issue we are facing by sticking with OpenFGA as persistent storage, same limitations apply to the ListObjects calls (not sure what's the max) and reason is that they are expensive calls

we should enable to choose those settings but at the same time move away from the List calls (and Read calls) and use our own database @nsklikas

shipperizer (approved):

PR #543, opened 2025-02-25
IAM-1336 Implement POC to test import from `identity-platform-api`

BarcoMasile:
## Description
This PR is only for demonstrating that the work done on Identity Platform API actually works as intended.

A Readme is provided to explain how to simply run the test server file and see the mock implementation in action.

**NB: As stated in the Readme** please know that after https://github.com/canonical/identity-platform-api/pull/4 the import path will change from "github.com/canonical/identity-platform-api/**gen**/roles" to "github.com/canonical/identity-platform-api/**v0**/roles".


test-server/entrypoint.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	// Start HTTP server (and proxy calls to gRPC server endpoint)
+	const listenAddress = ":8081"
+
+	signal := make(chan error)
+	go func() {
+		if err := http.ListenAndServe(listenAddress, mux); err != nil {
+			signal <- err
+			log.Println("ListenAndServe:", err)
```
wood-push-melon, line 88: LGTM. Just a dummy question:

is it better to do the logging before sending the error to the channel?

wood-push-melon (approved):

test-server/entrypoint.go
BarcoMasile: Yes, it would probably make more sense, I was conflicted over the order in which to perform these two actions
I thought this was the one but since you also prefer it that way you convinced me
I'll change it :)

BarcoMasile:
This was approved by @shipperizer .
I will close this PR and and complete the corresponding Jira task.

PR #544, opened 2025-02-26
feat: update go version to 1.24.0 + adapt code to changes

BarcoMasile:
## Description
Update go version to 1.23.6 and refactor code to comply with latest changes.
Subsequently update go version to 1.24.0 and also refactor code to comply with `vet` command complaints, more about this below.
This will allow us to keep up with latest Golang release more easily.

Right now this PR builds on top of https://github.com/canonical/identity-platform-admin-ui/pull/543 but commits will soon be sorted.

### About the `vet` command complaints
With go1.24 if you use `fmt.Errorf` with a string var as the first argument it complains and the build fails.
It has to be a string constant (which in Go is actually a constant string literal), that's why I initially used `fmt.Errorf("%s", stringVariable)`.
After Nikos comment I switched to using `errors.New(stringVar)` since it felt more appropriate.


internal/openfga/stores.go
```diff
@@ -279,7 +278,7 @@ func (s *OpenFGAStore) ListPermissions(ctx context.Context, ID string, continuat
 		eMsg = fmt.Sprintf("%s%v - %s\n", eMsg, n, e.Error())
 	}
 
-	return permissions, tMap, fmt.Errorf(eMsg)
+	return permissions, tMap, fmt.Errorf("%s", eMsg)
```
nsklikas, line 281: nit: I think you can use `errors.New`

nsklikas (approved):
lgtm. Only a very minor comment, feel free to ignore.

internal/openfga/stores.go
BarcoMasile: Yes thank you, actually that's what I wanted to use but Ale in the past has asked me to stick with fmt.Errorf.
I would like to change it to errors.New, do you approve @shipperizer ?

BarcoMasile: I pushed the change as a new commit in the meantime

nsklikas (approved):

internal/openfga/stores.go
nsklikas: I'm fine either way

BarcoMasile: BTW I just realized I didn't explain why I had to do this, I only mentioned the "vet command complaints".
Basically right now with go1.24 if you use `fmt.Errorf` with a string var it complains and the build fails.
It has to be a string constant (which in Go is actually a constant string literal).
I'll add this explanation to the description for context.

internal/tracing/tracer.go
```diff
@@ -19,12 +19,14 @@ import (
...
 	semconv "go.opentelemetry.io/otel/semconv/v1.18.0"
 	"go.opentelemetry.io/otel/trace"
+	"go.opentelemetry.io/otel/trace/embedded"
 
 	"github.com/canonical/identity-platform-admin-ui/internal/logging"
 )
 
 type Tracer struct {
 	tracer trace.Tracer
+	embedded.Tracer
```
shipperizer, line 29: mind sending a link to the change in the lib? The line above this seems redundant unless the two tracer interfaces are completely dislinked

rockcraft.yaml
```diff
@@ -29,7 +29,7 @@ parts:
     source: .
     source-type: local
     build-snaps:
-      - go/1.23/stable
+      - go/1.24/stable
```
shipperizer, line 32: ❤️

shipperizer (approved):
looks legit, just one question around the tracer, not arguing it's wrong but simply if you can save me from googling 

also I take fmt.errorf is put and errors.new is in, right?

internal/openfga/stores.go
shipperizer: right, my preference would be sticking with fmt.Errorf
sorry I read the pr description before the change and only came round to it now

anyway it is such a trivial thing that if you guys are happy with errors then let's stick to it

dunno if there is anything on the canonical golang style guide, but happy to go with whatever the team prefers

internal/tracing/tracer.go
BarcoMasile: Here's the [link](https://pkg.go.dev/go.opentelemetry.io/otel/trace#hdr-API_Implementations) to the paragraph.

TL;DR: this is due to the fact that the Otel library may present changes in the API even without a major version bump, so they use this golang "trick" (that I randomly learned last week from another library) to force you to notice the change with a compilation error.
The alternatives would be to not be alerted at all, or to be alerted with runtime panics.

So I chose this one to get alerted at compilation time. If you remove the embedeed struct we get compilation error because our object would not respect the interface.

nsklikas (approved):

PR #550, opened 2025-03-07
IAM-1311 Migrate roles api to gRPC-gateway based server

BarcoMasile:
## Description
This PR introduces the first API (Roles) based on the gRPC-gateway implementation from the generated code in https://github.com/canonical/identity-platform-api.

The gRPC_mapper.go file contains two functions needed to make sure the new APIs match the existing one.
- `SetHeaderFromMetadataFilter` is needed to make sure that response headers are set on the response object
- `ForwardErrorResponseRewriter` is needed to make sure error responses match the ones from the rest of the APIs

The gRPC_handlers.go contains the gRPC handlers implementations, mimicing what has been done for the handlers.go.

Test have been added for every new piece of code.

### Other changes
- `Response` object has been enriched with `omitempty` to avoid serializing nil values
- fixed a typo in the word "Rule" which is supposed to be "Role"

### DIFF
--- a/internal/http/types/gRPC_mappers.go
+++ b/internal/http/types/gRPC_mappers.go
@@ -1,0 +1,75 @@
+// Copyright 2025 Canonical Ltd.
+// SPDX-License-Identifier: AGPL-3.0
+
+package types
+
+import (
+	"context"
+	"net/http"
+	"strings"
+
+	"github.com/grpc-ecosystem/grpc-gateway/v2/runtime"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/metadata"
+	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/proto"
+
+	v0Types "github.com/canonical/identity-platform-api/v0/http"
+	rpcStatus "google.golang.org/genproto/googleapis/rpc/status"
+)
+
+// SetHeaderFromMetadataFilter is the filter function to allow headers from the handlers to be set
+// on the HTTP response by the gRPC gateway (to be registered with WithForwardResponseOption)
+// DO NOT use WriteHeader func in filters like this one
+//
+// usage example:
+//
+// mux := runtime.NewServeMux(
+//
+//	runtime.WithForwardResponseOption(SetHeaderFromMetadataFilter),
+//
+// )
+func SetHeaderFromMetadataFilter(ctx context.Context, w http.ResponseWriter, _ proto.Message) error {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return status.Errorf(codes.Internal, "error getting incoming context metadata")
+	}
+
+	for key, values := range md {
+		// filter out grpcgateway headers
+		if strings.HasPrefix(key, "grpcgateway-") {
+			continue
+		}
+
+		for _, v := range values {
+			w.Header().Add(key, v)
+		}
+	}
+
+	return nil
+}
+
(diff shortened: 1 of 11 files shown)


pkg/roles/gRPC_handlers.go
```diff
@@ -0,0 +1,305 @@
...
+	if role == nil || role.GetName() == "" {
+		return nil, status.Errorf(codes.InvalidArgument, "role name is empty")
+	}
+
+	principal := authentication.PrincipalFromContext(ctx)
+	createdRole, err := g.svc.CreateRole(ctx, principal.Identifier(), role.GetName())
+
+	if err != nil {
+		g.logger.Errorf("failed to create role: %v", err)
+		return nil, status.Errorf(codes.NotFound, "failed to create role: %v", err)
```
wood-push-melon, line 107: Just a question: why do we return `NotFound` when failing to create a role?

internal/http/types/token.go
```diff
@@ -64,6 +68,24 @@ func (p *TokenPaginator) LoadFromRequest(ctx context.Context, r *http.Request) e
...
+	_, span := p.tracer.Start(ctx, "types.TokenPaginator.LoadFromGRPCContext")
+	defer span.End()
+
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return nil
+	}
+
+	var header []string
+	if header, ok = md[GRPC_PAGINATION_METADATA]; !ok || header[0] == "" {
```
wood-push-melon, line 82: a minor question:

Do we need to check `len(header)` before `header[0] == ""`?

BarcoMasile: For incoming request, if the key is present than a value must be present in the slice, so in this case it should be safe to do so, but I will add the check just in case, you never know! :) Thanks!

BarcoMasile:
Addressed @wood-push-melon 's comments and rebased on main.

nsklikas (approved):
LGTM

pkg/web/router.go
```diff
@@ -32,6 +35,8 @@ import (
 	"github.com/canonical/identity-platform-admin-ui/pkg/rules"
 	"github.com/canonical/identity-platform-admin-ui/pkg/schemas"
 	"github.com/canonical/identity-platform-admin-ui/pkg/status"
+
+	//"github.com/canonical/identity-platform-admin-ui/pkg/status"
```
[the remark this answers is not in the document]
BarcoMasile: yes a leftover from my trials and erros, there are two packages with the same name and I had issues with importing the right one. Fixed this in the next PR I'm working on, thanks hawkeye 🦅

internal/http/types/gRPC_mappers.go
```diff
@@ -0,0 +1,75 @@
...
+// )
+func SetHeaderFromMetadataFilter(ctx context.Context, w http.ResponseWriter, _ proto.Message) error {
+	md, ok := metadata.FromIncomingContext(ctx)
+	if !ok {
+		return status.Errorf(codes.Internal, "error getting incoming context metadata")
+	}
+
+	for key, values := range md {
+		// filter out grpcgateway headers
+		if strings.HasPrefix(key, "grpcgateway-") {
```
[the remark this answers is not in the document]
BarcoMasile: I didn't put it as a const because it's used only here, but I'll look into relying on the lib constant (in the next PR I'm working on), thanks for the suggestion

PR #560, opened 2025-04-11
IAM-1315 Port identities api

BarcoMasile:
## Description
This PR add support for the Identities API (gRPC-gateway based).
A GrpcMapper struct has been added to brdige the mapping from and to the models used by admin uI, which are in turn taken from the Kratos SDK. This doesn't allow us to edit the json tags as I would have wanted to, that is why in the gRPC_mapper_test.go file you see I had to "normalize" some result since semantis is correctly preserved in the regular processing of the data by the actual service.

The gRPC mapper tests only focus on the happy path. This is due to difficulty in finding a way to generate and test an error that makes sense + the amount of time this task already took.

### Missing
gRPC_handlers_test file to cover for the five endpoints of the v0 Identities API.

### DIFF
--- a/pkg/identities/gRPC_handlers.go
+++ b/pkg/identities/gRPC_handlers.go
@@ -1,0 +1,232 @@
+// Copyright 2025 Canonical Ltd.
+// SPDX-License-Identifier: AGPL-3.0
+
+package identities
+
+import (
+	"context"
+	"net/http"
+	"strconv"
+
+	v0Types "github.com/canonical/identity-platform-api/v0/http"
+	"google.golang.org/grpc/codes"
+	"google.golang.org/grpc/status"
+
+	"github.com/canonical/identity-platform-admin-ui/internal/logging"
+	"github.com/canonical/identity-platform-admin-ui/internal/monitoring"
+	"github.com/canonical/identity-platform-admin-ui/internal/tracing"
+
+	v0Identities "github.com/canonical/identity-platform-api/v0/identities"
+)
+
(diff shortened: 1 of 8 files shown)


pkg/identities/gRPC_handlers.go
```diff
@@ -0,0 +1,232 @@
...
+		Status:  http.StatusOK,
+		Message: &message,
+	}, nil
+}
+
+func (g *GrpcHandler) CreateIdentity(ctx context.Context, req *v0Identities.CreateIdentityReq) (*v0Identities.CreateIdentityResp, error) {
+	ctx, span := g.tracer.Start(ctx, "identities.GrpcHandler.CreateIdentity")
+	defer span.End()
+
+	createIdentityBody := req.GetIdentity()
```
wood-push-melon, line 116: LGTM. A minor thing (I might be wrong, feel free to ignore):

Is it better to check the request content here instead of postponing to `g.svc.CreateIdentity`?

Same thing also happens in the `UpdateIdentity`.

wood-push-melon (approved):

pkg/identities/gRPC_handlers.go
BarcoMasile: Good question. Right this endpoints benefit from the same Validator processing of the http api, which stays in place for every endpoint despite the underlying implementation.
So right now all the payloads get validated before actually hitting the endpoint (whether http or grpc), and if the contents are not valid, the request is rejected.

BarcoMasile:
Rebased on main

PR #568, opened 2025-05-30
IAM-1540 Add database support

BarcoMasile:
## Description
This PR introduces PostgreSQL support with a DBClient providing an easy way to run queries with or without a surrounding database transaction.
DBClient also relies on a flag to possibly enable prepared statements cache, which is not manadatory (and needs to be turned off in case we rely on pg-bouncer for handling the connection pool to PostgreSQL).

When running queries in transactions there's full error recovery support, for both checked errors and for `panic`s.

Query cache cannot be used when running queries inside a transaction through Squirrel.
The PostgreSQL driver of choice is PGX, which by default always caches prepared statements. This makes it incompatible with PG Bouncer out of the box. Connection config can be changed to disable caching setting `DefaultQueryExecMode = pgx.QueryExecModeSimpleProtocol` but that would force us to do the interpolation of the query parameters ourselves, since that mode (the only one compatible with PGBouncer) also disables parameter interpolation.

In the future, if we decide to support PG Bouncer, we already have the native PGX connection available to set that config, but we will also have add an additional behavior to actually interpolate query parameters on our side with a simple additional function call. I considered PG Bouncer support out of scope for this PR since [IAM-1533](https://warthogs.atlassian.net/browse/IAM-1533).

## Libraries introduced
- PostgreSQL database driver - https://github.com/jackc/pgx
- Squirrel for query type safety - https://github.com/masterminds/squirrel
- SqlMock necessary to mock DB and Transactions - https://github.com/DATA-DOG/go-sqlmock

[IAM-1533]: https://warthogs.atlassian.net/browse/IAM-1533?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


pkg/storage/storage.go
```diff
@@ -0,0 +1,152 @@
...
+
+type QueryAction func(builderType sq.StatementBuilderType) (*ActionResult, error)
+type RollbackAction func(error)
+
+var (
+	NoopAction   QueryAction    = func(builderType sq.StatementBuilderType) (*ActionResult, error) { return nil, nil }
+	NoopRollback RollbackAction = func(error) {}
+)
+
+type ActionResult struct {
```
wood-push-melon, line 28: A dummy question:

Is it possible or idiomatic to consider using `interface` to separate the single and multiple cases?

BarcoMasile: I tried to find a common interface but there is not one, so I went with the simplest approach to implement kind of a "union" type structure, to allow for these 2 types that are possibly returned by the queries

wood-push-melon: Gotcha. So the caller will somehow use the `hasSingleRow` and `hasMultipleRows` to check the results?

Just a random thought (might not be a good idea, feel free to ignore): is it OK if we just keep `Rows` field in the `ActionResult`? Somehow, a single row is just a "special" case of multiple rows. And both implement `Scan`.

pkg/storage/storage.go
```diff
@@ -0,0 +1,152 @@
...
+	if err := tx.Commit(); err != nil {
+		d.logger.Errorf("Failed to commit transaction, err: %v", err)
+		return nil, err
+	}
+
+	return ret, nil
+}
+
+func NewDBClient(dsn string, queryCacheEnabled bool, tracer tracing.TracingInterface, monitor monitoring.MonitorInterface, logger logging.LoggerInterface) *DBClient {
+	config, err := pgx.ParseConfig(dsn)
```
wood-push-melon, line 132: Just a non-blocking random question (out of curiosity, not in this PR's scope):

Does the driver `pgx` support read/write traffic split natively? Or we still need to use something like HAProxy, etc.

BarcoMasile: Great question, I tested this driver with multiple postgresql at the same time and it works, although when we eventually add a load balancing implementation (whetehr hw or sw) we will have to rely on something else to distinguish between readonly queries and write queries. PGX officially does not support making that distinction.

pkg/storage/storage.go
BarcoMasile: Exactly. Technically the caller will also already know what to expect so it should not even need the check itself.
But at the same time I wanted to let the next tasks guide this part of the implementation.
About the Rows vs Scan, we could do that but then every result will have to be read in a for loop and results length would have to be checked. So I thought this would reduce boilerplate in retrieving results and streamlining (a bit) the caller implementation

pkg/storage/storage.go
```diff
@@ -0,0 +1,152 @@
...
+	ret, err := action(statementBuilder)
+	if err != nil {
+		d.logger.Errorf("Failed to execute QueryAction, err: %v", err)
+		return nil, err
+	}
+
+	return ret, nil
+}
+
+func (d *DBClient) RunInTransaction(ctx context.Context, action QueryAction, rollback RollbackAction) (*ActionResult, error) {
```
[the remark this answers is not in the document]
BarcoMasile: So this function is for reducing the boilerplate for running queries in a transaction (these lines of code would be otherwise repeated every time we need a tx), this way we have cleaner and streamlined implementations and have panic-proof code for rolling back our application logic no matter what happens.

I think with next steps it will be more clear how it's used and the benefit from it using the Squirrel object for running queries.

BarcoMasile:
> question(out of curiosity): did you consider sqlx/sqlc vs squirrel? They look like interesting alternatives, not saying that they would be better for the job, just wondering why we chose to go with squirrel.
> 
> thought: IMO using prepared statements sounds like an optimization and I think that it's too early to make such a decision. Also I think that it is likely that we will be using pgbouncer in a production env (?)

@nsklikas I know it may sound like optimization, but prepared statements are the default way to query a db, this is to prevent the db to compile over and over the same queries all the time, leading to much more cpu/memory consumption.
It's the same principle of function evaluation, the function is compiled and evaluated once, meanwhile parameters change and need to be evaluated every time. There are other optimization that could be applied (that I kept out of this for that reason you mentioned).

Sorry for the long answer about squirrel:
yes I considered alternatives (like gorm, sqlc), but I wanted to avoid ORMs and was close to choose sqlc but eventually went with squirrel for a couple of reasons:
- sqlc requires an extra step in the dev process, because you have to prepare the queries and then generate the code for them with the tool
- sqlc also didn't introduce any abstractions on top of the standard golang `database/sql` interface, which I tend to prefer if it makes sense
- gorm seems to have many issues and is not really recommended nowadays, also I wanted to avoid the complex layer of an ORM which would force us to optimize for the overhead of the orm itself
- squirrel is actively maintened, adds a thin layer on top of the standard library and allows for type safe queries at the same time, without requiring additional steps. Also, it allow us to use transactions in the abstraction layer allowing for reducing the boilerplate code that I mentioned earlier. This is the same library used by OpenFGA which was a nice coincidence, but didn't play a role in the choice

For pgbouncer I did some more diggin on Friday: it's useful for pooling connections in case we need multiple app connecting to our database, but it doesn't do load balancing (as I wrongfully remembered). In our case the db will only be accessed by Admin UI (for our current design) and integration with other apps will go through the Admin UI APIs. But of course if things change and we need pgbouncer, we can add support in no time, I did some digging on that.
We will need a load balancer in the future, either hardware or software, to support horizontally scaling PostgreSQL since its replicas would only support readonly queries, and started looking into that as well this past friday.

wood-push-melon (approved):
LGTM. Like the design :)

nsklikas (approved):

pkg/storage/storage.go
shipperizer: admin UI wont have traffic requiring this, so we should be fine

shipperizer, line 132: might wanna add a tracer lib for this https://github.com/exaring/otelpgx

PR #571, opened 2025-06-05
IAM-1539 Add goose migrations and related makefile targets

BarcoMasile:
## Description
This PR add the first goose migration and makefile targets to handle goose connection to the db, goose up and down commands.

N.B.: this initial schema is more than the Jira task requested. Reason is: it doesn't bother us to have the schemas already.

The most important change is the presence of both the `owner` and `owner_id` columns.
This is because we don't have a plan to integrate identities in our db yet.
So right now the Owner can just be a string referencing a Kratos identity.
By the end of this Epic we will hopefully have more clarity on this.

Some indices have been added, but those will be driven by our implementations of the queries we need, so they will be subject to addition or changes.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -58,3 +58,31 @@ dev:
 		--tail \
 		--cache-artifacts=false
 .PHONY: dev
+
+
+GOOSE=goose
+GOOSE_DRIVER=postgres
+GOOSE_DBSTRING?=postgresql://user:user@localhost:5432/admin-service?sslmode=disable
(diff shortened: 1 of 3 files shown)


shipperizer (commented):
Makefile wont' be part of the rock, so we need a couple of things:

* copy over the sql migrations (the folder) into the rock
* create a CLI command (same as the `version` or `server`) that calls goose, hopefully this works off the shelf without using the shell (as we won't have it)

PR #573, opened 2025-06-05
Disable kratos session on logout

natalian98:
### Changes summary
This PR introduces disabling kratos identity session on logout, once the Hydra session is deactivated.
The session id is first fetched from [`sessions/whoami`](https://www.ory.sh/docs/kratos/reference/api#tag/frontend/operation/toSession) endpoint by passing the kratos session cookie.

### Additional context
As pointed out by @nsklikas, Hydra might be able to propagate logout to Kratos when `urls.identity_provider.url` config is defined and `identity_provider_session_id` parameter is passed when [accepting](https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/acceptOAuth2LoginRequest) a login request (which we already do in Login UI), however we did not implement oidc logout in hydra but instead we just deactivate the session and clear cookies. For that reason we need to deactivate the kratos session on our own.

### Testing

Run the Login UI [docker-compose](https://github.com/canonical/identity-platform-login-ui/blob/main/README.md) and create a hydra client for Admin UI:
```
docker exec hydra \
  hydra create client \
    --endpoint http://127.0.0.1:4445 \
    --name admin-ui \
    --grant-type authorization_code,refresh_token \
    --response-type token,code,id_token \
    --format json \
    --scope openid,offline_access,profile,email \
    --redirect-uri http://localhost:8888/api/v0/auth/callback
```

Then update it with audience set to the generated client id (we could possibly set the client id on creation but I didn't test that):
```
docker exec hydra \
    hydra update client <client-id> \
    --skip-tls-verify --name admin-ui --secret <client-secret> \
    --skip-consent --grant-type authorization_code,refresh_token \
    --response-type token,code,id_token  \
    --scope openid,offline_access,profile,email \
    --redirect-uri http://localhost:8888/api/v0/auth/callback \
    --endpoint http://localhost:4445 \
    --audience <client-id>
```

Go to openfga (:3000/playground) and create a model with the correct [schema](https://github.com/canonical/identity-platform-admin-ui/blob/cbc5f539bc3a68c00eb03c0b4cc38f43de4e3e15/internal/authorization/schema.openfga#L4). Retrieve the store and model id from database or logs.

Set the environment variables:
```
export ACCESS_TOKEN_VERIFICATION_STRATEGY=jwks
export AUTHENTICATION_ENABLED=true
export AUTHORIZATION_ENABLED=false
export CONTEXT_PATH=/
export HYDRA_ADMIN_URL=http://localhost:4445
export IDP_CONFIGMAP_NAME=idps
export IDP_CONFIGMAP_NAMESPACE=default
export KRATOS_ADMIN_URL=http://localhost:4434
export KRATOS_PUBLIC_URL=http://localhost:4433
export KUBECONFIG_FILE=~/.kube/config
export LOG_LEVEL=DEBUG
export MAIL_FROM_ADDRESS="Identity Platform <test@example.com>"
export MAIL_HOST=localhost
export MAIL_PASSWORD=test
export MAIL_PORT=1026
export MAIL_USERNAME=
export OATHKEEPER_PUBLIC_URL=http://localhost:4456
export OAUTH2_AUTH_COOKIES_ENCRYPTION_KEY=WrfOcYmVBwyduEbKYTUhO4X7XVaOQ1wF
export OAUTH2_CLIENT_ID=client-id
export OAUTH2_CLIENT_SECRET=client-secret
export OAUTH2_CODEGRANT_SCOPES=openid,profile,email,offline_access
export OAUTH2_REDIRECT_URI=http://localhost:8888/api/v0/auth/callback
export OIDC_ISSUER=http://localhost:4444
export OPENFGA_API_HOST=localhost:8080
export OPENFGA_API_SCHEME=http
export OPENFGA_API_TOKEN=42
export OPENFGA_AUTHORIZATION_MODEL_ID=openfga-model-id
export OPENFGA_STORE_ID=openfga-store-id
export PORT=8888
export RULES_CONFIGMAP_NAME=oathkeeper-rules
export RULES_CONFIGMAP_NAMESPACE=default
export SCHEMAS_CONFIGMAP_NAME=identity-schemas
export SCHEMAS_CONFIGMAP_NAMESPACE=default
export TRACING_ENABLED=false
```
Build with `go build -o admin-ui` and run. Log in at `http://localhost:8888/api/v0/auth`, then log out `/api/v0/auth/logout`.

### DIFF
--- a/pkg/authentication/handlers_test.go
+++ b/pkg/authentication/handlers_test.go
@@ -14,6 +14,7 @@ import (
 	"testing"
 
 	"github.com/coreos/go-oidc/v3/oidc"
+	kClient "github.com/ory/kratos-client-go"
 	"go.opentelemetry.io/otel/trace"
 	"go.uber.org/mock/gomock"
 	"golang.org/x/oauth2"
@@ -40,6 +41,7 @@ func TestHandleLogin(t *testing.T) {
 		Times(1).
 		Return(context.TODO(), trace.SpanFromContext(context.TODO()))
 	mockMonitor := NewMockMonitorInterface(ctrl)
+	mockSessionManager := NewMockSessionManagerInterface(ctrl)
 
 	mockHelper := NewMockOAuth2HelperInterface(ctrl)
 	mockHelper.EXPECT().RandomURLString().Return("mock-nonce")
@@ -67,6 +69,7 @@ func TestHandleLogin(t *testing.T) {
 		NewOAuth2Context(config, mockOIDCProviderSupplier(&oidc.Provider{}, nil), mockTracer, mockLogger, mockMonitor),
 		mockHelper,
 		NewAuthCookieManager(mockTTLSeconds, mockTTLSeconds, mockEncrypt, mockLogger),
+		mockSessionManager,
 		mockTracer,
 		mockLogger,
 	)
@@ -108,6 +111,7 @@ func TestHandleLoginCallback(t *testing.T) {
 	mockLogger := NewMockLoggerInterface(ctrl)
 	mockLogger.EXPECT().Debugf(gomock.Any(), gomock.Any()).Times(1)
 	mockTracer := NewMockTracer(ctrl)
+	mockSessionManager := NewMockSessionManagerInterface(ctrl)
 
 	mockHelper := NewMockOAuth2HelperInterface(ctrl)
 	mockEncrypt := NewMockEncryptInterface(ctrl)
@@ -148,6 +152,7 @@ func TestHandleLoginCallback(t *testing.T) {
 		mockOauth2Ctx,
 		mockHelper,
 		NewAuthCookieManager(mockTTLSeconds, mockTTLSeconds, mockEncrypt, mockLogger),
+		mockSessionManager,
 		mockTracer,
 		mockLogger,
 	)
@@ -187,6 +192,7 @@ func TestHandleLoginCallbackFailures(t *testing.T) {
 	defer ctrl.Finish()
 
 	mockTracer := NewMockTracer(ctrl)
+	mockSessionManager := NewMockSessionManagerInterface(ctrl)
 
 	mockToken := &oauth2.Token{}
 	mockToken.AccessToken = "mock-access-token"
@@ -355,6 +361,7 @@ func TestHandleLoginCallbackFailures(t *testing.T) {
 				mockOauth2Ctx,
 				mockHelper,
 				NewAuthCookieManager(mockTTLSeconds, mockTTLSeconds, mockEncrypt, mockLogger),
+				mockSessionManager,
 				mockTracer,
 				mockLogger,
 			)
@@ -400,6 +407,7 @@ func TestHandleMe(t *testing.T) {
 	mockTracer := NewMockTracer(ctrl)
 	mockOauth2Ctx := NewMockOAuth2ContextInterface(ctrl)
 	mockLogger := NewMockLoggerInterface(ctrl)
+	mockSessionManager := NewMockSessionManagerInterface(ctrl)
 
 	mockHelper := NewMockOAuth2HelperInterface(ctrl)
 	mockEncrypt := NewMockEncryptInterface(ctrl)
@@ -411,6 +419,7 @@ func TestHandleMe(t *testing.T) {
 		mockOauth2Ctx,
 		mockHelper,
 		NewAuthCookieManager(mockTTLSeconds, mockTTLSeconds, mockEncrypt, mockLogger),
+		mockSessionManager,
 		mockTracer,
 		mockLogger,
 	)
@@ -453,15 +462,19 @@ func TestLogout(t *testing.T) {
 	for _, tt := range []struct {
 		name         string
 		errorMessage string
-		setupMocks   func(*MockOAuth2ContextInterface, *MockAuthCookieManagerInterface, *MockLoggerInterface)
+		setupMocks   func(*MockOAuth2ContextInterface, *MockAuthCookieManagerInterface, *MockLoggerInterface, *MockSessionManagerInterface)
 	}{
 		{
 			name:         "Success",
 			errorMessage: "",
-			setupMocks: func(c *MockOAuth2ContextInterface, m *MockAuthCookieManagerInterface, l *MockLoggerInterface) {
+			setupMocks: func(c *MockOAuth2ContextInterface, m *MockAuthCookieManagerInterface, l *MockLoggerInterface, s *MockSessionManagerInterface) {
 				m.EXPECT().ClearIDTokenCookie(gomock.Any())
 				m.EXPECT().ClearAccessTokenCookie(gomock.Any())
 				m.EXPECT().ClearRefreshTokenCookie(gomock.Any())
+
+				sessionID := "test"
+				s.EXPECT().GetIdentitySession(gomock.Any(), gomock.Any()).Return(&SessionData{Session: kClient.Session{Id: sessionID}}, nil)
+				s.EXPECT().DisableSession(gomock.Any(), sessionID).Return(&SessionData{Session: kClient.Session{}}, nil)
 
 				c.EXPECT().Logout(gomock.Any(), gomock.Any()).Times(1).Return(nil)
 			},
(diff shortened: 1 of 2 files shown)


pkg/authentication/handlers.go
```diff
@@ -253,6 +254,26 @@ func (a *API) handleLogout(w http.ResponseWriter, r *http.Request) {
...
+	session, err := a.identities.GetIdentitySession(ctx, r.Cookies())
+	if err != nil {
+		a.logger.Errorf("failed to retrieve kratos session, err: %v", err)
+		a.badRequest(w, err)
+		return
+	}
+	if session == nil {
+		a.badRequest(w, fmt.Errorf("failed to get session id"))
+		return
+	}
```
BarcoMasile, line 266: nitpick: feel free to disregard, the cases handled here seem to be pretty much equivalent, we could just sum it up in one single case (we can still print the error if it happens to be nil), for example (take this only as an example)

```suggestion
	if err != nil || session == nil {
	        err = fmt.Errorf("failed to retrieve kratos session or session not found, err: %v", err)
		a.logger.Error(err.Error())
		a.badRequest(w, err)
		return
	}
```

pkg/authentication/handlers.go
```diff
@@ -253,6 +254,26 @@ func (a *API) handleLogout(w http.ResponseWriter, r *http.Request) {
...
+
+	sessionID := session.GetSession().Id
+	_, err = a.identities.DisableSession(ctx, sessionID)
+
+	if err != nil {
+		a.logger.Errorf("failed to disable kratos session, err: %v", err)
+		a.badRequest(w, err)
+		return
```
BarcoMasile, line 274: nitpick: it would possibly be better to have the same error returned to the user in the logs, so
```suggestion
                err = fmt.Errorf("failed to disable kratos session, err: %v", err)
		a.logger.Error(err.Error())
		a.badRequest(w, err)
		return
```

pkg/identities/service.go
```diff
@@ -76,6 +91,59 @@ func (s *Service) parseError(r *http.Response) *kClient.GenericError {
 	return gerr.Error
 }
 
+func (s *Service) cookiesToString(cookies []*http.Cookie) string {
+	var ret = make([]string, len(cookies))
+	for i, c := range cookies {
+		ret[i] = fmt.Sprintf("%s=%s", c.Name, c.Value)
+	}
+	return strings.Join(ret, "; ")
+}
```
BarcoMasile, line 100: question: since cookies is a slice it could potentially be `nil`. Are the arguments passed to this function possibly nil?

pkg/authentication/handlers.go
natalian98: I splitted those in case session is nil but there is no error, I'd prefer to keep it

pkg/authentication/handlers.go
```diff
@@ -253,6 +254,27 @@ func (a *API) handleLogout(w http.ResponseWriter, r *http.Request) {
...
+		a.badRequest(w, fmt.Errorf("failed to get session id"))
+		return
+	}
+
+	sessionID := session.GetSession().Id
+	_, err = a.identities.DisableSession(ctx, sessionID)
+
+	if err != nil {
+		err = fmt.Errorf("failed to disable kratos session, err: %v", err)
+		a.logger.Error(err.Error())
```
wood-push-melon, line 273: A minor thing:

I guess it's supposed to be something like `a.logger.Errorf("failed to disable kratos session, err: %v", err)`.

pkg/identities/service.go
```diff
@@ -76,6 +91,59 @@ func (s *Service) parseError(r *http.Response) *kClient.GenericError {
...
+
+	// GET /sessions/whoami
+	session, rr, err := s.kratosPublic.
+		ToSession(ctx).
+		Cookie(s.cookiesToString(cookies)).
+		Execute()
+
+	data := new(SessionData)
+	if err != nil {
+		s.logger.Error("failed to get identity session: %v", err)
```
wood-push-melon, line 114: Should be `Errorf`?

pkg/authentication/interfaces.go
```diff
@@ -112,3 +114,13 @@ type PrincipalInterface interface {
 	RefreshToken() string
 	IDToken() string
 }
+
+type KratosSession interface {
```
wood-push-melon, line 118: Have a question about this interface:

Why do we need this interface? In the `GetIdentitySession` and `DisableSession`, we could just return the `SessionData`, right? And we'd better not return an interface (restrict on what it returns).

pkg/identities/handlers.go
```diff
@@ -45,8 +45,9 @@ func (a *API) RegisterEndpoints(mux *chi.Mux) {
 	mux.Put("/api/v0/identities/{id:.+}", a.handleUpdate)
 	// mux.Patch("/api/v0/identities/{id:.+}", a.handlePartialUpdate)
 	mux.Delete("/api/v0/identities/{id:.+}", a.handleRemove)
+	mux.Delete("/api/v0/identities/sessions/{id:.+}", a.handleSessionDisable)
```
wood-push-melon, line 48: A minor question: Is this API for compensation?

pkg/authentication/interfaces.go
natalian98: Returning `SessionData` was my initial idea, but I needed to work around a circular dependency issue: `authentication` package is [imported](https://github.com/canonical/identity-platform-admin-ui/blob/ff8880d519d56907133d2f7039748e0e0d39628e/pkg/identities/gRPC_handlers_test.go#L22) in `identity`, and to return `SessionData` I'd have to import `identity` in `authentication`, which introduces the circular dependency.

pkg/identities/handlers.go
BarcoMasile: Question: we want to disable kratos session **on logout** right? So the logic to do that (which is now in "handleSessionDisable") should be part of the logout process, not a new API endpoint. That's what I was expecting, but I may be missing pieces

shipperizer:
> As pointed out by @nsklikas, Hydra might be able to propagate logout to Kratos when urls.identity_provider.url config is defined and identity_provider_session_id parameter is passed when [accepting](https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/acceptOAuth2LoginRequest) a login request (which we already do in Login UI), however we did not implement oidc logout in hydra but instead we just deactivate the session and clear cookies. For that reason we need to deactivate the kratos session on our own.

i'm interested on this point here, is there scope to reuse hydra functionality? cc  @nsklikas

nsklikas (commented):
approved on accident...

pkg/identities/handlers.go
shipperizer:
```suggestion
	mux.Delete("/api/v0/identities/{id:.+}sessions/{sid:..+}", a.handleSessionDisable)
	mux.Delete("/api/v0/identities/{id:.+}sessions", a.handleSessionDisableAll)
```

for these endpoints you need to add related authorization permissions
ideally we need the caller to have access to the user with `can_view` and `can_edit`

pkg/identities/handlers.go
```diff
@@ -264,6 +265,30 @@ func (a *API) handleRemove(w http.ResponseWriter, r *http.Request) {
 	)
 }
 
+func (a *API) handleSessionDisable(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
+	sessionID := chi.URLParam(r, "id")
+
+	session, err := a.service.DisableSession(r.Context(), sessionID)
```
shipperizer, line 272:
```suggestion
	session, err := a.service.DisableSession(r.Context(), chi.URLParam(r, "id"))
```

pkg/authentication/handlers.go
```diff
@@ -73,6 +73,7 @@ type API struct {
 	oauth2           OAuth2ContextInterface
 	helper           OAuth2HelperInterface
 	cookieManager    AuthCookieManagerInterface
+	identities       SessionManagerInterface
```
shipperizer, line 76: naming seems a bit off for now?maybe sessions or sessionManager?

```suggestion
	sessions       SessionManagerInterface
```

shipperizer:
> > > As pointed out by @nsklikas, Hydra might be able to propagate logout to Kratos when urls.identity_provider.url config is defined and identity_provider_session_id parameter is passed when [accepting](https://www.ory.sh/docs/hydra/reference/api#tag/oAuth2/operation/acceptOAuth2LoginRequest) a login request (which we already do in Login UI), however we did not implement oidc logout in hydra but instead we just deactivate the session and clear cookies. For that reason we need to deactivate the kratos session on our own.
> > 
> > 
> > i'm interested on this point here, is there scope to reuse hydra functionality? cc @nsklikas
> 
> Yes, but not only that:
> 
> * The admin UI would be making 1 HTTP call (hydra would be making 1 call to kratos to logout) rather than 3
> * It would allow other applications to logout from kratos as well, as we wouldn't be using the internal hydra/kratos APIs
> * Our implementation wouldn't rely in the kratos cookie being available (imo we should limit the kratos/hydra cookie only to specific paths, so that they are not accessible by all apps on our domain)
> 
> The downside with that approach is that it would require some changes to the login UI as well (implement a couple more endpoints) and maybe some more careful design.

right, seems a decent amount of work, potentially requiring a bit of speccing and design, shall we make an issue?considering this in the long term scenario of having a shared session across our internal services (hook, admin, login, verification) can this help us getting there?

pkg/authentication/session.go
```diff
@@ -0,0 +1,111 @@
...
+
+type KratosError struct {
+	Error *kClient.GenericError `json:"error,omitempty"`
+}
+
+func (s *SessionData) GetError() *kClient.GenericError {
+	return s.Error
+}
+
+func (s *SessionData) GetSession() kClient.Session {
```
wood-push-melon, line 41: A minor thing:

Seems we don't need these two `getter`s anymore if we don't do any specific handling for getting the fields (and also the fields are exported.)

pkg/authentication/session.go
```diff
@@ -0,0 +1,111 @@
...
+		ToSession(ctx).
+		Cookie(s.cookiesToString(cookies)).
+		Execute()
+
+	data := new(SessionData)
+	if err != nil {
+		s.logger.Errorf("failed to get identity session: %v", err)
+		// data.Error = s.parseError(rr)
+		data.Error = kratos.ParseKratosError(rr)
+		return data, err
```
wood-push-melon, line 68: A minor thing:

Seems that the caller `handleLogout` does not handle the specific Kratos error. If this is the case, we could directly return `nil` without parsing the Kratos error.

natalian98: I think there is still value in returning the kratos error, even if the `handleLogout` implementation doesn't process it.
This service may be reused by other packages, and preserving it will provide better context for future error handling.

BarcoMasile (dismissed):

nsklikas (commented):
overall lgtm, only a couple of nit picks/questions

pkg/authentication/handlers.go
```diff
@@ -253,6 +254,23 @@ func (a *API) handleLogout(w http.ResponseWriter, r *http.Request) {
 	a.cookieManager.ClearAccessTokenCookie(w)
 	a.cookieManager.ClearRefreshTokenCookie(w)
 
+	s, err := a.sessions.GetIdentitySession(ctx, r.Cookies())
+	if err != nil {
+		a.logger.Errorf("failed to retrieve kratos session, err: %v", err)
+		a.badRequest(w, err)
```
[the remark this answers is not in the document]
natalian98: 400 is a standard: https://www.ory.sh/docs/kratos/reference/api#tag/identity/operation/disableSession

pkg/kratos/errors.go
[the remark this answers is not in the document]
natalian98: You are right, I moved it to the internal kratos package.

pkg/authentication/session.go
```diff
@@ -0,0 +1,95 @@
...
+	// GET /sessions/whoami
+	session, rr, err := s.kratosPublic.
+		ToSession(ctx).
+		Cookie(s.cookiesToString(cookies)).
+		Execute()
+
+	data := new(SessionData)
+	if err != nil {
+		s.logger.Errorf("failed to get identity session: %v", err)
+		data.Error = kratos.ParseKratosError(rr)
```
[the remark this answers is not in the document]
natalian98: Yes, the sdk error returns just the response status (e.g. `400 Bad Request`), while the error response also contains the kratos error message (e.g. "uuid: incorrect UUID length 35 in string"). While I don't think it makes sense to expose the complete error details to the user, I added the kratos error message to the logs.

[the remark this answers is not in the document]
nsklikas: I have to say I don't really like that we have to introduce a new struct and that we return 2 errors from the same function. But I guess there are limitations from the sdk and it's not major.

wood-push-melon (approved):

PR #574, opened 2025-06-06
IAM-1538 Add CLI command to perform database migrations

BarcoMasile:
## Changes
This PR adds the CLI command to perform database migrations (up, down) and migrations status check.
These commands directly map to [Goose](https://github.com/pressly/goose) commands
- `migrate [up]` => `goose up`
- `migrate down` => `goose down` (it downgrades by one single migration)
- `migrate down 0` => `goose down-to 0` (downgrades removing all migrations effects)
- `migrate down 2` => `goose down-to 2` (downgrades down to version 2)
- `migrate status` => `goose status`

**NB**: Makefile db-down command still removes all migrations for ease of development

A new custom validation function has been added to validate the parameters and a unit test has been added in order to make sure that the migrate CLI command is consistent.

Other changes are:
- refactored the makefile target db-down to perform full rollback instead of one single step down
- added `EmbedMigrations embed.FS` in `migrations` folder (since it couldn't be put in the CLI folder since it would require to embed a folder path starting with `../` which is not supported by `embed`
- added Ping check to make sure connection to the DB is actually working (this was supposed to be performed by the PGX driver but it turned out not true)
- reordered migration deletions (down) to avoid dependencies


nsklikas (approved):
lgtm

cmd/migrate.go
```diff
@@ -0,0 +1,76 @@
...
+	goose.SetBaseFS(migrations.EmbedMigrations)
+
+	if err := goose.SetDialect("postgres"); err != nil {
+		return err
+	}
+
+	switch command {
+	case "up":
+		err = goose.UpContext(ctx, db, ".", goose.WithNoColor(true))
+	case "down":
```
nsklikas, line 69: thought: Does this clean the database? Is this what we want? Shouldn't we allow the user to choose what migration to downgrade to?

pkg/storage/storage.go
```diff
@@ -135,6 +135,9 @@ func NewDBClient(dsn string, queryCacheEnabled bool, tracer tracing.TracingInter
 	}
 
 	db := stdlib.OpenDB(*config)
+	if err := db.Ping(); err != nil {
```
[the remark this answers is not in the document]
BarcoMasile: We are in the first case, we would still be able to connect to the database, but when a query is executed, if the query is selecting comuns that are no longer there, then we would get a runtime error.

This should not happen, as the charm should always do the migration which is idempotent in case all the available migrations were already performed. Migrations are embedded in the executable so the app will always have "the right ones".

Adding a validation of the schema is something that we could look into, not a bad idea at all!
Thanks!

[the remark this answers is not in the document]
BarcoMasile: https://github.com/canonical/identity-platform-admin-ui/issues/577

cmd/migrate.go
```diff
@@ -0,0 +1,76 @@
...
+	migrateCmd.Flags().String("dsn", "", "PostgreSQL DSN connection string")
+	_ = migrateCmd.MarkFlagRequired("dsn")
+
+	rootCmd.AddCommand(migrateCmd)
+}
+
+func migrate(ctx context.Context, dsn, command string) error {
+	config, err := pgx.ParseConfig(dsn)
+	if err != nil {
+		return fmt.Errorf("DSN validation failed, shutting down, dsn: %s, err: %v", dsn, err)
```
wood-push-melon, line 52: A minor point: better not put the DSN value in (username and password included).

BarcoMasile: You're right, didn't think about that, thanks!

cmd/migrate.go
```diff
@@ -0,0 +1,76 @@
...
+	}
+
+	db := stdlib.OpenDB(*config)
```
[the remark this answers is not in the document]
BarcoMasile: I relied on the official doc that says:
> The returned [DB](https://pkg.go.dev/database/sql#DB) is safe for concurrent use by multiple goroutines and maintains its own pool of idle connections. Thus, the OpenDB function should be called just once. It is rarely necessary to close a [DB](https://pkg.go.dev/database/sql#DB).

So I didn't close it. I could expose a close func on the struct anyway so we can close the db anyway. I'll do it

[link](https://pkg.go.dev/database/sql#Open:~:text=The%20returned%20DB%20is%20safe%20for%20concurrent%20use%20by%20multiple%20goroutines%20and%20maintains%20its%20own%20pool%20of%20idle%20connections.%20Thus%2C%20the%20OpenDB%20function%20should%20be%20called%20just%20once.%20It%20is%20rarely%20necessary%20to%20close%20a%20DB.)

wood-push-melon (dismissed):

shipperizer (approved):

cmd/migrate.go
BarcoMasile: Ok I will refactor this and implement this behavior. Thank you both for the exchange!

shipperizer (approved):

nsklikas (approved):

PR #584, opened 2025-06-12
IAM-1529 Move roles CRUD operations to database

BarcoMasile:
## Description
This PR moves CRUD operations from OpenFGA to the database.
It also adds the necessary env var to handle the db connection, called Data Source Name (DSN).

This PR also adjust the initial goose migration to remove the owner related fields from the Policy table since policies are "owned" by the role they are attached to and their lifecycle is attached to the role itself, so in the end the owner is the same owner of the role.

The new RoleRepository has been adopted in the Service implementation for Roles for CRUD operations.
In the storage package, the DBClient has been simplified to skip the "RunInTransaction" method and to simply pass provide a Squirrel StatementBuilderType with a new transactions, letting the caller of the Repository method to decide if the Transactions needs to be commited or rolled back.

Tests for the service have been also refactore but they do not consider all paths for the repository method results given the already complicated structure of the current tests that will need to be refactored

The file CONTRIBUTING.md has been updated with some basic info on using `sqlmock` for unit tests, since many issues have been encountered while using it.

Repository methods have been uniformed in the parameters list so that role reference parameter comes first (right after the context).


pkg/storage/interfaces.go
```diff
@@ -0,0 +1,16 @@
...
+import (
+	"context"
+
+	sq "github.com/Masterminds/squirrel"
+)
+
+type DBClientInterface interface {
+	Statement() sq.StatementBuilderType
+	RunInTransaction(ctx context.Context, action QueryAction, rollback RollbackAction) (*ActionResult, error)
+	Close() error
```
wood-push-melon, line 15: Have a minor question about the `Close()`:

Are we going to use it somewhere? How are we handling the graceful shutdown?

And I came across the following documentation in `sqlx`:

```
// OpenDBFromPool creates a new *sql.DB from the given *pgxpool.Pool. Note that this method automatically sets the
// maximum number of idle connections in *sql.DB to zero, since they must be managed from the *pgxpool.Pool. This is
// required to avoid acquiring all the connections from the pgxpool and starving any direct users of the pgxpool. Note
// that closing the returned *sql.DB will not close the *pgxpool.Pool.
func OpenDBFromPool(pool *pgxpool.Pool, opts ...OptionOpenDB) *sql.DB {
	c := GetPoolConnector(pool, opts...)
	db := sql.OpenDB(c)
	db.SetMaxIdleConns(0)
	return db
}
```

(I might be understanding it wrong, feel free to ignore it): `sql.DB` is a wrapper of `pgxpool.Pool` based on the `NewDBClient` logic. Do we need to actually close the `pool` instead of `db` in the `Close()` implementation?

pkg/roles/repository.go
```diff
@@ -0,0 +1,170 @@
...
+
+type RoleRepository struct {
+	db storage.DBClientInterface
+
+	logger  logging.LoggerInterface
+	tracer  tracing.TracingInterface
+	monitor monitoring.MonitorInterface
+}
+
+func (r *RoleRepository) FindRole(ctx context.Context, userID, ID string) (*Role, error) {
```
wood-push-melon, line 28: A minor thing:

The `ID` input is not used in this method.

pkg/roles/repository.go
```diff
@@ -0,0 +1,170 @@
...
+		QueryRowContext(ctx)
+
+	id, name := "", ""
+
+	err := row.Scan(&id, &name)
+	if err != nil {
+		return nil, fmt.Errorf("unable to scan FindRole result, %w", err)
+	}
+
+	if id == "" {
```
wood-push-melon, line 44: A minor thing:

Probably align with the `ListRoles`? like:

```
if errors.Is(err, sql.ErrNoRows) {
    return nil, storage.ErrNotFound
}

if err != nil {
    return nil, fmt.Errorf("unable to scan FindRole result, %w", err)
}
```

pkg/roles/repository.go
```diff
@@ -0,0 +1,170 @@
...
+func (r *RoleRepository) FindRole(ctx context.Context, userID, ID string) (*Role, error) {
+	ctx, span := r.tracer.Start(ctx, "roles.RoleRepository.GetRole")
+	defer span.End()
+
+	row := r.db.Statement().Select("id", "name").
+		From(storage.ROLE_TABLE).
+		Where(sq.Eq{"owner": userID}).
+		QueryRowContext(ctx)
+
+	id, name := "", ""
```
wood-push-melon, line 37: (cosmetics, feel free to ignore):

```
var id, name string
```

pkg/roles/repository.go
```diff
@@ -0,0 +1,170 @@
...
+func (r *RoleRepository) ListRoles(ctx context.Context, userID string) ([]string, error) {
+	ctx, span := r.tracer.Start(ctx, "roles.RoleRepository.ListRoles")
+	defer span.End()
+
+	rows, err := r.db.Statement().Select("name").
+		From(storage.ROLE_TABLE).
+		Where(sq.Eq{"owner": userID}).
+		Query()
+
+	if errors.Is(err, sql.ErrNoRows) {
```
wood-push-melon, line 63: (Disregard it if I'm wrong):

It seems `Query` does not return such error `sql.ErrNoRows`.

pkg/roles/repository.go
```diff
@@ -0,0 +1,170 @@
...
+		Suffix("RETURNING id").
+		QueryRowContext(ctx)
+
+	var deletedId int64 = -1
+	err := row.Scan(&deletedId)
+	if err != nil {
+		return 0, err
+	}
+
+	if deletedId == -1 {
```
wood-push-melon, line 153: Similar here. Not sure if we could rely on the error returned by `Scan` by checking:

```
if errors.Is(err, sql.ErrNoRows) {
	return 0, storage.ErrNotFound
}
if err != nil {
    return 0, err
}
```

pkg/roles/repository.go
```diff
@@ -0,0 +1,170 @@
...
+
+	return &role, nil
+}
+
+func (r *RoleRepository) DeleteRole(ctx context.Context, userID, roleID string) (int64, error) {
+	ctx, span := r.tracer.Start(ctx, "roles.RoleRepository.DeleteRole")
+	defer span.End()
+
+	row := r.db.Statement().Delete(storage.ROLE_TABLE).
+		Where(sq.Eq{"owner": userID}).
```
wood-push-melon, line 142: 👍

pkg/roles/repository.go
```diff
@@ -0,0 +1,170 @@
...
+	monitor monitoring.MonitorInterface
+}
+
+func (r *RoleRepository) FindRole(ctx context.Context, userID, ID string) (*Role, error) {
+	ctx, span := r.tracer.Start(ctx, "roles.RoleRepository.GetRole")
+	defer span.End()
+
+	row := r.db.Statement().Select("id", "name").
+		From(storage.ROLE_TABLE).
+		Where(sq.Eq{"owner": userID}).
```
wood-push-melon, line 34: If this is a frequent access pattern, I guess we'd better change the index in the `00002_add_intial_indices.sql` from

```
CREATE INDEX idx_role_name_owner ON role (name, owner);
```

to

```
CREATE INDEX idx_role_name_owner ON role (owner, name);
```

for better use the index that follows the leftmost prefix rule.

pkg/roles/repository.go
BarcoMasile: you're right, in my trial & error I left some code in shambles and will fix it asap, sorry

pkg/storage/interfaces.go
BarcoMasile: I added the close method becuase later in development I'd like to see a fit for it, although from what I gathered online it's not really crucial to handle db connection closure (as it is in Java or other languages/frameworks).

The settings of the connection (max idle connections, max and min overall connections, etc..) is something that will be tackled in the future as part of the "tuning" of the DB pool, which is too soon to tackle right now, but which is the exact reason why I still used the "native" pgx client to then convert to the standard library wrapper database/sql interface (which handles a pool itself).

The alternative was to juse use the `sql.Open("postgres")` method which under the hood uses pgx driver, but doesn't give us access to the connection settings that we will use.

TL;DR: for now I wouldn't bother with db close, but will come as part of the tuning of the tuning of the pool connections

pkg/roles/repository.go
BarcoMasile: I was not able to confirm this. Techinally the Query method is part of the Squirrel library, which in turn invokes methods of the underlying database/sql interface, which should throw that error.
I am currently testing these behavior in particular and should be able to provide more info a little later. I will avoid putting the PR in draft for this and hopefully have something more clear by the end of today

pkg/roles/repository.go
BarcoMasile: the order should not matter as long as an index is available for the column requested. Anyway the right order is the one you see from the original index, I fixed the implementation and will push now to address your comments (which I appreciated a lot, thank you)

pkg/roles/repository.go
```diff
@@ -0,0 +1,171 @@
...
+
+	if errors.Is(err, sql.ErrNoRows) {
+		return []string{}, nil
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("unable to list roles, %w", err)
+	}
+
+	defer func() { _ = rows.Close() }()
```
nsklikas, line 72: (probably stupid) question: can't we re-write this to
```suggestion
	defer rows.Close()
```
?

pkg/roles/repository.go
wood-push-melon: Gotcha. My knowledge about index comes from MySQL, probably not completed-suited to Postgres.

BarcoMasile: Just to be super clear, usually the order of columns in the definition/creation of the idex **matters**.
The order of the "WHERE" conditions don't affect the usage of the index.
So there is definitely truth to what you wrote :)

pkg/storage/interfaces.go
BarcoMasile: EDIT: I added the db Close invocation in the end :)

pkg/roles/repository.go
BarcoMasile: Finally I was able to confirm: you are 100% correct, thank you for pointing me in the right direction
Here's the [explanation](https://pkg.go.dev/database/sql#ErrNoRows:~:text=var%20ErrNoRows%20%3D,until%20a%20Scan.)

pkg/roles/repository.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	return &Role{
+		ID:    id,
+		Name:  name,
+		Owner: owner,
+	}, nil
+}
+
+func (r *RoleRepository) FindRoleByNameAndOwner(ctx context.Context, name string, userID string) (*Role, error) {
```
shipperizer, line 83: any reason to not use a variadic arg with the option pattern, something like withName or withOwner?

nice to have potentially, not a deal breaker

pkg/roles/service.go
```diff
@@ -77,56 +77,66 @@ func (s *Service) GetRole(ctx context.Context, userID, ID string) (*Role, error)
...
+	createdRole, tx, err := s.repo.CreateRoleTx(ctx, userID, roleName)
+	if err != nil {
+		err = fmt.Errorf("unable to create role %s for user %s, %w", roleName, userID, err)
+		s.logger.Error(err.Error())
+		return nil, err
+	}
 
-	// TODO @shipperizer offload to privileged creator object
-	role := fmt.Sprintf("role:%s", ID)
+	role := fmt.Sprintf("role:%s", roleName)
```
shipperizer, line 109: isn't it better sticking with the id in OpenFGA too?see the complication of not having the id until that is returned by the dB, but wondering if this is wise in light of competing names (even though we should have a unique index on name?)

pkg/roles/repository.go
```diff
@@ -0,0 +1,273 @@
...
+
+	var id, _name, owner string
+
+	err := row.Scan(&id, &_name, &owner)
+	if errors.Is(err, sql.ErrNoRows) {
+		return nil, storage.ErrNotFound
+	}
+
+	if err != nil {
+		return nil, fmt.Errorf("unable to scan FindRoleByIdAndOwner result, %w", err)
```
wood-push-melon, line 101: A minor typo: `FindRoleByNameAndOwner`.

pkg/roles/repository.go
```diff
@@ -0,0 +1,273 @@
...
+
+type RoleRepository struct {
+	db storage.DBClientInterface
+
+	logger  logging.LoggerInterface
+	tracer  tracing.TracingInterface
+	monitor monitoring.MonitorInterface
+}
+
+func (r *RoleRepository) FindRoleById(ctx context.Context, ID string) (*Role, error) {
```
wood-push-melon, line 28: nonblocking, feel free to ignore:

Seems that there is a common logic in the three role-finding methods. Could be extracting it.

pkg/storage/storage.go
```diff
@@ -58,83 +52,29 @@ type DBClient struct {
...
-	if action == nil {
-		d.logger.Error("query action cannot be null")
-		return nil, fmt.Errorf("query action cannot be null")
-	}
-
-	if rollback == nil {
-		rollback = NoopRollback
-	}
-
+func (d *DBClient) TxStatement(ctx context.Context) (TxInterface, sq.StatementBuilderType, error) {
```
wood-push-melon, line 61: A bit concerned here (I might be pedantic, feel free to ignore):

Why do we return an interface? The producer usually returns a concrete type, and the consumer can declare an interface with necessary methods.

pkg/roles/repository.go
```diff
@@ -0,0 +1,273 @@
...
+}
+
+func (r *RoleRepository) CreateRole(ctx context.Context, userID, roleName string) (*Role, error) {
+	ctx, span := r.tracer.Start(ctx, "roles.RoleRepository.CreateRole")
+	defer span.End()
+
+	return createRoleQuery(ctx, r.db.Statement(), userID, roleName)
+}
+
+func (r *RoleRepository) CreateRoleTx(ctx context.Context, userID, roleName string) (*Role, storage.TxInterface, error) {
```
wood-push-melon, line 183: Nonblocking, feel free to ignore:

A bit concerned that we keep popping tx to the service layer, and let it control the tx commit and rollback (understand the point that we would like to control the cross-system tx).

Originally, I thought db tx was only targeted for multiple db operations. Personally prefer the original implementation. But also good with the current one.

pkg/roles/service.go
```diff
@@ -77,56 +77,66 @@ func (s *Service) GetRole(ctx context.Context, userID, ID string) (*Role, error)
...
 	user := fmt.Sprintf("user:%s", userID)
 
-	err := s.ofga.WriteTuples(
-		ctx,
+	tuples := []ofga.Tuple{
```
wood-push-melon, line 112: Just a random thought:

Another thing we might want to pay attention later if we are using this pattern in other resources (like groups) is that:

We might not want to do something with undetermined latency while holding the write lock (for example, when updating a record concurrently).

pkg/roles/service.go
BarcoMasile: Synced offline

pkg/roles/repository.go
BarcoMasile: it would be nice but since either 1 or 2 methods of these will be most likely removed after it wouldn't be worth it

pkg/roles/repository.go
BarcoMasile: I understand your concern, the reason for giving the tx control to the service is to let service rollback in case openfga fails, so it's a necessary control to make sure the service can implement the design and rely on the success of the openfga operation. The previous implementation was a little cumbersome especially due to the strict typing system Go uses.

pkg/roles/service.go
```diff
@@ -77,56 +77,66 @@ func (s *Service) GetRole(ctx context.Context, userID, ID string) (*Role, error)
...
 		*ofga.NewTuple(user, authorization.ASSIGNEE_RELATION, role),
 		*ofga.NewTuple(user, authorization.CAN_VIEW_RELATION, role),
-	)
+		*ofga.NewTuple(user, authorization.CAN_EDIT, role),
+		*ofga.NewTuple(user, authorization.CAN_DELETE, role),
```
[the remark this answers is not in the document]
BarcoMasile: you're absolutely right, I'll change this!

pkg/roles/repository.go
```diff
@@ -0,0 +1,273 @@
...
+func (r *RoleRepository) ListRoles(ctx context.Context, userID string, page, size int64) ([]string, error) {
+	ctx, span := r.tracer.Start(ctx, "roles.RoleRepository.ListRoles")
+	defer span.End()
+
+	pageSize := storage.PageSize(size)
+	offset := storage.Offset(page, pageSize)
+
+	rows, err := r.db.Statement().Select("name").
+		From("role").
+		Where(sq.Eq{"owner": userID}).
```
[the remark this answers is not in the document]
BarcoMasile: Yes it will change with the introduction of a new relation on openfga that will allow a user to manage resources of other users. In the spec I wrote an example with a permission called `manageGroupMembersResources` to give this idea.
When all is completed, the ownership will allow owners to handle their resources without needing extra tuples on openfga.
A lot will change (authz, this implementation also) to support the full spec.

PR #589, opened 2025-06-20
IAM-1276 Improve NewRouter function and streamline code

BarcoMasile:
## Description
This PR refactors the NewRouter function to use the functional options pattern, allowing for cleaner and more flexible configuration. The following changes were made:

### Changes
- adopted functional options pattern in NewRouter for better extensibility and readability
- simplified logic in NewRouter, reducing repetition and unnecessary
- Removed infrequently used variables that were only declared and used once, inlined their usage instead
- Removed the ui.Config struct for its simplicity, but we should add it back as soon as the UI handler needs more info

## Small consideration on the functional options pattern
If we like this, we could easily apply the same pattern to our API handlers and services.
The "missing" part to make this actually 100% effective is to have sensible defaults for every settings in the available options, which right now we don't have.
Right now we have to pass **all** needed options to make sure the Router works as expected because the current defaults wouldn't allow the app to work with "missing" options compared to the ones we pass right now. Do we need it strictly? No. Would it allow us to streamline the code even more? Yes.

@shipperizer

### DIFF
--- a/cmd/serve.go
+++ b/cmd/serve.go
@@ -34,7 +34,6 @@ import (
 	"github.com/canonical/identity-platform-admin-ui/pkg/rules"
 	"github.com/canonical/identity-platform-admin-ui/pkg/schemas"
 	"github.com/canonical/identity-platform-admin-ui/pkg/storage"
-	"github.com/canonical/identity-platform-admin-ui/pkg/ui"
 	"github.com/canonical/identity-platform-admin-ui/pkg/web"
 )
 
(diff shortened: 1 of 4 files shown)


nsklikas (dismissed):

wood-push-melon (dismissed):
lgtm.

(out of the scope of this PR) have a gut feeling about the factory method of `Router`: it is getting bigger.

shipperizer (dismissed):
broadly happy with this

re the defaults, ideally we have a "WithDefaults" config that gets initialized in the beginning, with NoOps structs, then with the options we fill it accordingly
NoOp is mainly to avoid optional stuff failing when not needed

shipperizer:
let's drop the Rules pkg in one separate commit, codebase invcluded

BarcoMasile:
Update: rebased on main and removed the changes related to the Rules package. It will come in a different PR in #592

PR #594, opened 2025-06-30
IAM-1609 Remove `rules` from project

### LINKED ISSUE #592: Drop the Rules package and related codebase
We won't be supporting Rules API anymore due to a chance in component support

the Rules API have become obsolete and we should remove it from the codebase

BarcoMasile:
CC @huwshimi : I don't see any reference to the Rules API from the frontend code, so I don't think we need any more changes.

Closes #592


nsklikas (approved):

wood-push-melon (approved):
We could probably drop the oathkeeper deployment in the local testing setup, structure test, and docs too. Happy to do it in a separate PR.

PR #613, opened 2025-07-24
IAM-1638 Support multiple API versions

BarcoMasile:
## Description
This PR separates different version of the API in different routers, so that multiple API versions are effectively supported.

## Changes
- OpenFGA related interfaces are now centralized in one single more readable interface
- original authz model for OpenFGA has been renamed to reflect the API version ti belongs to
- moving `web` package interfaces to `config` package
- added v0 openfga model (not used right now)
- the openfga model handling has been refactored into a Provider object

### DIFF
--- a/internal/authorization/auth_model.go
+++ b/internal/authorization/auth_model.go
@@ -70,11 +70,3 @@ func NewAuthorizationModelProvider(apiVersion string) *AuthorizationModelProvide
 
 	return a
 }
-
-var V0AuthzModelProvider *AuthorizationModelProvider = nil
-var V2AuthzModelProvider *AuthorizationModelProvider = nil
-
-func init() {
-	V0AuthzModelProvider = NewAuthorizationModelProvider("v0")
-	V2AuthzModelProvider = NewAuthorizationModelProvider("v2")
-}
--- a/cmd/createFgaModel.go
+++ b/cmd/createFgaModel.go
@@ -89,7 +89,8 @@ func createModel(apiUrl, apiToken, storeId, kubeconfig, k8sConfigMap string) {
 		fgaClient.SetStoreID(ctx, storeId)
 	}
 
-	authzModel := authorization.V0AuthzModelProvider.GetModel()
+	authzModel := authorization.NewAuthorizationModelProvider("v0").
+		GetModel()
 
 	modelId, err := fgaClient.WriteModel(
 		context.Background(),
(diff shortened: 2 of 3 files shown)


BarcoMasile:
Rebased on main

shipperizer (commented):
I'm not against this design but I would have kept all at the API level, aka each package implemented a NewAPI with the options pattern, in there we add our v0 and v2 seamlessly 

this kinda allows us to go up to different versions per endpoint domain, aka v3 roles and v4 groups while rest starts in v2, and we don't need to change much apart from those 2 pkgs

if the whole solution was done to cater for the rebac v1 case, I do see some benefit in standardising around that pattern but I'm not fully sold on it

wood-push-melon:
LGTM on the code. I am cool with both designs depending on what we need (I probably lack some context).

internal/authorization/auth_model.go
```diff
@@ -13,14 +13,20 @@ import (
...
-//go:embed schema.openfga
-var schema string
+//go:embed authorization_model.v0.openfga
+var v0Schema string
 
-// Taken from
-// https://github.com/openfga/cli/blob/d5bfb08cd540dc7c10737bcda12dbc292a649e22/internal/authorizationmodel/model.go#L156
-var AuthModel = func() openfga.AuthorizationModel {
+//go:embed authorization_model.v2.openfga
+var v2Schema string
```
nsklikas, line 20: dummy question: we don't have a `v1Schema` because it's the same as v0?

pkg/groups/service_test.go
```diff
@@ -36,6 +36,7 @@ import (
 
 //go:generate mockgen -build_flags=--mod=mod -package groups -destination ./mock_logger.go -source=../../internal/logging/interfaces.go
 //go:generate mockgen -build_flags=--mod=mod -package groups -destination ./mock_interfaces.go -source=./interfaces.go
+//go:generate mockgen -build_flags=--mod=mod -package groups -destination ./mock_openfga_interfaces.go -source=../../internal/openfga/interfaces.go
```
nsklikas, line 39: question: is this needed?

nsklikas (approved):
lgtm, a couple of dummy questions out of curiosity

internal/authorization/auth_model.go
```diff
@@ -34,5 +40,33 @@ var AuthModel = func() openfga.AuthorizationModel {
...
+	var model string
+	switch a.apiVersion {
+	case "v2":
+		model = v2Schema
+	default:
+		model = v0Schema
+	}
+
+	return readAuthzModelFromDSLString(model)
+}
```
shipperizer, line 57: if we have 2 schemas how do we deal with migrations?are the OpenFGA struct methods ready for dealing with separate models?
what is the new schema and how does it change from v0?

shipperizer (approved):

PR #655, opened 2025-09-17
fix: various startup bugs

andogq:
While trying to get the platform running, I've come across a number of issues which I've fixed:

- Add missing DSN env var
- Create `admin-service` database along with other databases
- Use `iam:iam` as postgres credentials for goose migrations
- `idps` -> `providers` for `IDPCONFIG_NAME`
- JSON typo in default IDP configuration

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -62,7 +62,7 @@ dev:
 
 GOOSE=goose
 GOOSE_DRIVER=postgres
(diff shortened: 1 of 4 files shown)


BarcoMasile (approved):
Thank you for the fixes, they are much appreciated!

nsklikas (approved):

